PR #50, opened 2021-02-22
Line ref

domfournier:
Release 0.3.1

- Fix numpy warnings for deprecated conversion to numpy.float
- Add checks for selection of a zero length line in Peak Finder.
- Allow line selection based on ReferencedData.
- Add base run test for all apps


docs/content/installation.rst
```diff
@@ -30,16 +30,20 @@ Installation
 	    :width: 50%
 
 
-.. note:: The assumption is made here that Anaconda has been installed in the default directory
+.. attention:: The assumption is made that Anaconda has been installed in the default directory:
+        ``%USERPROFILE%\anaconda3``
```
sebhmg, line 34: do you really need the full Anaconda, or would it work with miniconda?

sebhmg: I should try on my computer with `C:\Users\sebastienh\AppData\Local\Continuum\miniconda3` ...

geoapps/processing/peak_finder.py
```diff
@@ -2014,9 +2030,19 @@ def get_line_indices(self, line_id):
...
-            np.asarray(self.survey.get_data(self.lines.data.value)[0].values) == line_id
-        )[0]
+        line_data = self.survey.get_data(self.lines.data.value)[0]
+
+        if isinstance(line_data, ReferencedData):
+            line_id = [
+                key
+                for key, value in line_data.value_map.map.items()
+                if value == line_id
+            ]
```
sebhmg, line 2040: **suggestion:** extract as a function.

same logic repeated in em1d_inversion.py and plotting.py

Maybe a member function on `ReferencedData`?

geoapps/plotting/plotting.py
```diff
@@ -230,8 +230,16 @@ def plot_plan_data_selection(entity, data, **kwargs):
             if not np.any(entity.get_data(key)):
                 continue
 
+            line_data = entity.get_data(key)[0]
+            if isinstance(line_data, ReferencedData):
+                values = [
+                    key
+                    for key, value in line_data.value_map.map.items()
+                    if value in values
+                ]
```
sebhmg, line 239: **question:** not sure to understand why we need filtering if ReferenceData but not in the other cases

geoapps/processing/peak_finder.py
```diff
@@ -1071,12 +1073,12 @@ def objects_change(self, _):
...
-                            not_tem = False
+                            # not_tem = False
                             break
 
-            self.tem_checkbox.unobserve(self.objects_change, names="value")
-            self.tem_checkbox.value = not_tem is False
-            self.tem_checkbox.observe(self.objects_change, names="value")
+            # self.tem_checkbox.unobserve(self.objects_change, names="value")
+            # self.tem_checkbox.value = not_tem is False
+            # self.tem_checkbox.observe(self.objects_change, names="value")
```
sebhmg, line 1081: **suggestion:** remove lines that are commented out

docs/content/installation.rst
domfournier: would be great. And we could try with Miniconda for sure. It would be a change for current users though

geoapps/processing/peak_finder.py
```diff
@@ -2176,7 +2202,8 @@ def reset_groups(self, gates={}):
 
                 except IndexError:
                     print(
-                        "Could not find a time channel for the given list of time channels"
+                        "Could not find a time channel for the given list of time channels."
+                        + "Switching to non-tem mode"
```
sebhmg, line 2206: **suggestion (non-blocking, style):** preferred way to break a long string literal
```suggestion
                        ("Could not find a time channel for the given list of time channels. "
                         "Switching to non-tem mode.")
```

Concatenation is done at compile time (while `+`  does it at runtime)

geoapps/processing/peak_finder.py
```diff
@@ -2200,22 +2227,32 @@ def reset_groups(self, gates={}):
...
+            )
             d_min = np.min([d_min, channel["values"].min()])
             d_max = np.max([d_max, channel["values"].max()])
 
         if d_max > -np.inf:
             self.plot_trigger.value = False
             self.min_value.value = d_min
-        # self.set_default_groups(self.channels.options)
+            self.scale_value.value = thresh_value
+            # self.set_default_groups(self.channels.options)
```
sebhmg, line 2255: **suggestion:** remove line commented out

sebhmg (changes requested):
looks good 👍 
Please see some suggestions/questions

geoapps/plotting/plotting.py
domfournier: Want to expose the string values to the users, but do the logical indexing based on the keys (integer). I agree that a get_key from value would make sense.

I would suggest we leave it as is for now, create the function on geoh5py, and clean this up here after?

sebhmg: yep, could create a technical story in JIRA ;)
so that when this is done in geoh5py, it is remembered to update the code here

geoapps/processing/peak_finder.py
```diff
@@ -2202,8 +2202,8 @@ def reset_groups(self, gates={}):
 
                 except IndexError:
                     print(
-                        "Could not find a time channel for the given list of time channels."
-                        + "Switching to non-tem mode"
+                        "Could not find a time channel for the given list of time channels.",
+                        "Switching to non-tem mode",
```
sebhmg, line 2206: **suggestion:** use string concatenation

here it uses multiple print parameters (still runtime, while string concatenation with `()` syntax is compile time).
```suggestion
                        ("Could not find a time channel for the given list of time channels. "
                        "Switching to non-tem mode")
```

geoapps/processing/peak_finder.py
sebhmg: still a bunch of commented lines here

sebhmg (changes requested):
just 2 little things remaining

geoapps/processing/peak_finder.py
```diff
@@ -2202,8 +2188,8 @@ def reset_groups(self, gates={}):
 
                 except IndexError:
                     print(
-                        "Could not find a time channel for the given list of time channels.",
-                        "Switching to non-tem mode",
+                        "Could not find a time channel for the given list of time channels."
+                        "Switching to non-tem mode"
```
sebhmg, line 2192:
```suggestion
                        ("Could not find a time channel for the given list of time channels. "
                        "Switching to non-tem mode.")
```

PR #51, opened 2021-02-23
Setup precommit

sebhmg:
install precommit in your local clone with:
```sh
pre-commit install
pre-commit install -t pre-push
```

See also [pre-commit documentation](https://pre-commit.com/#install).


.github/workflows/pytest-windows.yaml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Get full Python version
+        id: full-python-version
+        shell: bash
+        run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
+      - name: Install dependencies with conda
+        run: |
+          C:\Miniconda\condabin\conda.bat env update --file environment.yml --name base
```
sebhmg, line 26: installing packages from environment.yml takes too long for a Github action.
@domfournier, could you have another file (e.g. `environment-test.yml`) which lists only the required packages for tests?

domfournier: hmmm might be tough since the apps depend on so many libraries... not sure what we would be testing without'em

sebhmg: I would assume unit tests do not require notebook, ipy modules  nor matplotlib or plotly

domfournier: fair enough, let me see where I can cut

domfournier: actually that's tricky since we import those libraries in the __init__

PR #52, opened 2021-03-02
Handle case where rotation is not set in JSON input file.

orerocks:
Fix abort when rotation is not set in the JSON input.


domfournier (approved):
Good catch. Thanks for this.

PR #53, opened 2021-03-03
Fix inconsistencies in magnetic inversion types. 

orerocks:
A reference to `mvic' was present after the assert statement would have failed. Different references to mvi or magnetics or mvic in different places: sometimes three types being set as vector mag (mvi, magnetics, mvic), and sometime two types (mvi, magnetics).

I streamlined all the calls with: magnetics( = magsus), mvi (spherical), and mvic (cartesian), and eliminated mag. I think you were using magnetics to mean MVIC, but I thought it made sense to clarify which was a vector and which wasn't.


domfournier:
@orerocks can you try on your side with the `mvic` flag. Last commit should fix it.

domfournier (approved):
Looking good on this side.

orerocks:
Yep, MVIC working as it should. Thanks!

PR #64, opened 2021-03-30
Geopy 110

benk-mira:


geoapps/io/driver.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class InputFile:
+    def __init__(self, filename):
+        self._accepted_file_extensions = ["json"]
+        self.data = None
+        self.filename = filename
+
+    @property
+    def filename(self):
```
domfournier, line 19:
```suggestion
    def filename(self) -> str:
```

geoapps/io/driver.py
```diff
@@ -0,0 +1,39 @@
...
+        return self._filename
+
+    @filename.setter
+    def filename(self, f):
```
domfournier, line 23:
```suggestion
    def filename(self, f: str):
```

geoapps/io/driver.py
```diff
@@ -0,0 +1,39 @@
...
+#
+#  geoapps is distributed under the terms and conditions of the MIT License
+#  (see LICENSE file at the root of this source code package).
+
+import json
+import os
+
+
+class InputFile:
+    def __init__(self, filename):
```
domfournier, line 13:
```suggestion
    def __init__(self, filename: str):
```

tests/input_file_test.py
```diff
@@ -0,0 +1,35 @@
...
+import json
+import os
+
+import pytest
+
+from geoapps.io import InputFile
+
+
+def test_create_work_path():
+    input_file = InputFile("../assets/test.json")
```
domfournier, line 17: That's great. Maybe we can store this json inside the test geoh5 file once the test is written.

domfournier (commented):
Looking solid bro. If you are ok with this, maybe we can start using `typing` for all our inputs and attributes output?

geoapps/io/driver.py
benk-mira: Just noticed these comments now.  I went back over and added typing support, but I didn't think to do it for the setters/getters.

tests/input_file_test.py
benk-mira: I've changed the way I tested in the latest commits, if I wrote any actual files to disk it uses tmp_path and shouldn't persist on disk after the testing.

geoapps/io/constants.py
```diff
@@ -0,0 +1,143 @@
...
+    "upper_bound",
+    "max_distance",
+    "max_chunk_size",
+    "chunk_by_rows",
+    "output_tile_files",
+    "no_data_value",
+    "parallelized",
+]
+
+required_parameters = ["inversion_type", "core_cell_size"]
```
domfournier, line 67: `required_parameters = ["inversion_type", ]`

geoapps/io/constants.py
```diff
@@ -0,0 +1,143 @@
...
+
+valid_parameter_values = {
+    "inversion_type": ["gravity", "magnetics", "mvi", "mvic"],
+    "inversion_style": ["voxel"],
+    "data_format": ["ubc_grav", "ubc_mag", "GA_object"],
+    "inversion_mesh_type": ["TREE"],
+    "gradient_type": ["total"],
```
domfournier, line 74: `"gradient_type": ["total", "components"]`

geoapps/io/validators.py
```diff
@@ -0,0 +1,268 @@
...
+                v = input["reference_model"]["value"]
+                v = np.r_[v]
+                if v.shape[0] not in [1, 3]:
+                    msg = "Start model needs to be a scalar or 3 component vector"
+                    raise ValueError(msg)
+        elif "save_to_geoh5" in input.keys():
+            if "out_group" not in input.keys():
+                msg = "Input file must contain a 'out_group' string if"
+                msg += "'save_to_geoh5' provided."
+                raise ValueError(msg)
```
domfournier, line 112: Could we condense all of the above with a list of required keys attached to each property?

geoapps/io/validators.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def validate_input(self, input: Dict[str, Any]) -> None:
+        """Validates input params and contents/type/shape/keys of values.
+
+        Calls validate method on individual key/value pairs in input, and
+        handles validations requiring knowledge of other parameters.
+
+        Parameters
+        ----------
+        input : dict
```
domfournier, line 61: I think @sebhmg would prefer to use the more compact version of `:param name: Decription`. The `type` isn't required technically since already defined above with typing.

domfournier: Same would have to change for all the other functions.

geoapps/io/constants.py
```diff
@@ -0,0 +1,143 @@
...
+
+valid_parameter_keys = {
+    "window": ["center_x", "center_y", "width", "height", "azimuth", "center", "size"],
+    "data": ["type", "name", "channels"],
+    "topography": ["GA_object", "drapped", "constant", "file"],
+    "receivers_offset": ["constant", "constant_drape", "radar_drape"],
+    "reference_model": ["model", "value", "none"],
+    "starting_model": ["model", "value"],
+    "depth_core": ["value", "auto"],
+}
```
domfournier, line 143: Would you consider changing the storage format as 
`"parameter_X": {
    "type": "val",
    "shape": (*, ),
    "required": True,
    "accepted": ["val_A", "val_B"]
}`
Then the whole validation could be done in a single loop?

domfournier (changes requested):
Fine work brother. See comments below.

geoapps/io/validators.py
sebhmg: true

geoapps/io/constants.py
benk-mira: sure thing!

geoapps/io/validators.py
benk-mira: Ok, I'll switch formats.  Good point about the typing.

geoapps/io/constants.py
benk-mira: will do.

geoapps/io/constants.py
benk-mira: Of course, since we're moving to providing the mesh!

benk-mira:
Latest push should fix up all the requested changes.

domfournier:
Are the test running locally?
image

geoapps/io/constants.py
```diff
@@ -0,0 +1,256 @@
+#  Copyright (c) 2021 Mira Geoscience Ltd.
+#
+#  This file is part of geoapps.
+#
+#  geoapps is distributed under the terms and conditions of the MIT License
+#  (see LICENSE file at the root of this source code package).
+
+
+required_parameters = ["inversion_type"]
```
domfournier, line 9: I would think that `data`, `topography` and `mesh` would be also base requirements

geoapps/io/validators.py
```diff
@@ -0,0 +1,244 @@
...
+        Raises
+        ------
+        ValueError, TypeError KeyError whenever an input parameter fails one of
+        it's value/type/shape/requirement validations.
+        """
+
+        self._validate_required_parameters(input)
+
+        if "data" in input.keys():
+            require_keys = "type" not in input["data"].keys()
```
domfournier, line 66: A bit like we did for the mesh, I am tempted to enforce that `Data` can only come from geoh5. It's just odd to have an exception for data. Ideally, the InputValidator should be very general to any application.

geoapps/io/validators.py
```diff
@@ -0,0 +1,244 @@
...
+            require_keys |= "name" not in input["data"].keys()
+            if require_keys:
+                raise KeyError("Data 'type' and 'name' must not be empty.")
+        elif "reference_model" in input.keys():
+            if "value" in input["reference_model"].keys():
```
domfournier, line 71: Same here, I think we want to generalize `reference_model` and assume that only a vector, float or None is provided. To simplify this let's split the `vector` case into 3 distinct reference models (amplitude, inclination, declination). Same would technically apply to starting model also... just not implemented here.

geoapps/io/validators.py
domfournier: Looks like `"type"` (no s) is only used by `Data` anyway right? It might simplify the checks...

domfournier (commented):
We are really close. I think if we just streamline some of the inversion parameters we can make the Validators completely generic, and avoid having special if statements for fringe cases...

domfournier (approved):
Woot! woot!

domfournier (approved):
Yes, again.

PR #65, opened 2021-03-30
Patch 0 4 1

domfournier:
One big long laundry list of updates. Don't do this at home kids.

- Improve docstrings + typing of utils functions
- Change LineDataDerivatives to use simple difference instead of FFT
- Fixes to data interpolation


domfournier:
@sebhmg Would you mind taking a look at the pytest fail? Something brakes when installing conda?

domfournier:
@sebhmg Ok so removing 3.7 from the list of version appears to do the trick. It was failing on figuring out `python[version=3.7.*, 3.8.*]` ?

PR #79, opened 2021-04-30
Geopy 177 alpha

domfournier:
**GEOPY-177 - python version not used in github actions** 
 Tester for 177


domfournier:
For testing GEOPY-177

PR #85, opened 2021-05-18
GEOPY-167 Add string_name formatter utils

domfournier:
Remove special characters by default from name of new objects. 
Defaults to replacing "." to "_"


benk-mira (approved):
Looks fine to me.  The string_name function is straight forward and the test looks good.  If you're satisfied that there's no cases where the '.' -> '_' transformation is undesirable then this works fine.  I guess the code that puts to '.' into the filenames is buried in GA somewhere, hey?

Note: you can iterate over a string without list conversion first: for char in characters should do the same as for char in list(characters).

domfournier:
right the "." happens when you convert float values to string.

PR #88, opened 2021-05-21
Update octree_mesh_creation.ipynb

RichardScottOZ:
Couple of typos.


domfournier:
Sorry @RichardScottOZ, we did similar edits at the same time. I will bring this in with #89

RichardScottOZ:
Noticing them is the important thing.

PR #89, opened 2021-05-25
GEOPY-149

domfournier:
**GEOPY-149 - Create octree mesh creation application** 
 - Add docs for the octree mesh creation app.
- Minor bug fixes


benk-mira (dismissed):
Nice work, I like having a bit of the algorithm details in the doc!  Should help people understand how to use the parameters better!

PR #90, opened 2021-05-26
Update create_octree.ipynb

RichardScottOZ:
minor edits


domfournier:
Same here, I had to do the edits manually, but big thanks for reporting!

PR #103, opened 2021-06-09
Geopy 131

benk-mira:
**GEOPY-131 - Unit test: octree_2_treemesh** 
 tests octree_2_treemesh for all combinations of corner origins and negative cell sizes


domfournier (approved):
Cool, did not know about the itertools.product. Looks like we might be repeating some of the combinations but that's fine.

PR #105, opened 2021-06-10
GEOPY-247

domfournier:
**GEOPY-247 - Speed up tests** 
 Speed up tests


benk-mira (approved):
All good! all the waiting is happening in the MVIParams constructors when it creates a workspace from path.  If we opened up the constructor to take a workspace object as a parameter we could just pass the object to all the tests and they would probably run in much less time.

tests/params_test.py
```diff
@@ -145,12 +145,14 @@ def param_test_generator(tmp_path, param, value):
     ui[param]["enabled"] = True
```
benk-mira, line 145: If you add a ui["geoh5"] = None here _init_params will bypass another time-consuming Workspace creation.

benk-mira (approved):
Merge if you're happy with my changes

PR #110, opened 2021-06-18
GEOPY-252: InversionData class 

benk-mira:
**GEOPY-252 - create a survey inversion component class** 
 Initial design finished with tests passing locally.  Opening the pull request early to see if the tests will pass in github actions.


domfournier (approved):
Looking good. Thanks for the test again

PR #112, opened 2021-06-21
GEOPY-251 : topography class

benk-mira:
**GEOPY-251 - Create a topography inversion component class** 
 Abstracted the locations windowing, filtering from the data class to an InversionLocations class and inherited to support the topography class.


geoapps/drivers/components/data.py
```diff
@@ -0,0 +1,248 @@
...
+    def _initialize(self):
+        """ Extract data from params class. """
+
+        self.ignore_value, self.ignore_type = self.parse_ignore_values()
+        self.components, self.data, self.uncertainties = self.get_data()
+
+        self.locs = super().get_locs(self.params.data_object)
+        self.mask = np.ones(len(self.locs), dtype=bool)
+
+        if self.window is not None:
```
domfournier, line 48: I m wondering if this should be moved to the __init__ of InversionLocation since common to topo and data

geoapps/drivers/components/data.py
```diff
@@ -0,0 +1,248 @@
...
+            self.mask = filter_xy(
+                self.locs[:, 0],
+                self.locs[:, 1],
+                window=self.window,
+                angle=self.angle,
+                mask=self.mask,
+            )
+
+        if self.params.resolution is not None:
```
domfournier, line 57: Same here

geoapps/drivers/components/locations.py
```diff
@@ -0,0 +1,83 @@
...
+
+from uuid import UUID
+
+import numpy as np
+from geoh5py.objects import Grid2D
+
+from geoapps.utils import rotate_xy
+
+
+class InversionLocations:
```
domfournier, line 16: This is a great idea. See comment above to consolidate more of the core xyz operations.

geoapps/drivers/components/locations.py
```diff
@@ -0,0 +1,83 @@
...
+        :returns: locs: N by 3 array of x, y, z location data
+
+        """
+
+        data_object = self.workspace.get_entity(uid)[0]
+
+        if isinstance(data_object, Grid2D):
+            locs = data_object.centroids
+        else:
+            locs = data_object.vertices
```
domfournier, line 69: This might break weird if the user inputs something else than a Grid2D. Maybe add a check after for `locs is not None`?

domfournier (commented):
Looking good. See comments

geoapps/drivers/components/data.py
benk-mira: get_locs needs the topo/data specific attribute in self.params so the locs call can't get moved up the hierarchy.  If the locs can't then neither can the mask.

geoapps/drivers/components/data.py
benk-mira: Not sure which chunk you're referring to, but the resolution isn't shared by both topo and data, since I leave the topo sampling as is.  And the windowing is handled different in each too (topo copies the window and expands it's x/y limits

geoapps/drivers/components/data.py
domfournier: Agreed

geoapps/drivers/components/data.py
domfournier: Makes sense

geoapps/drivers/components/locations.py
benk-mira: Fixed up.  BlockModel initializes with vertices=None as you suspected.

domfournier (approved):
Wunderbar

PR #118, opened 2021-06-30
Geopy 269

benk-mira:
**GEOPY-269 - drape option in base inversion driver not working** 
 Added a z_from_topo ui element to receivers group.  When True, the receivers z location is built by interpolation into the topography.  This occurs outside the drape method during class initialization and the drape method now just takes a radar offset array and applies the z translation.  Tests updated.  Inversions are now putting the data where it should be.

### DIFF
--- a/geoapps/drivers/components/data.py
+++ b/geoapps/drivers/components/data.py
@@ -9,7 +9,6 @@ from copy import deepcopy
 from typing import Dict, Tuple
 
 import numpy as np
-from scipy.interpolate import LinearNDInterpolator
 from SimPEG import maps
 from SimPEG.utils.drivers import create_nested_mesh
 
@@ -77,6 +76,8 @@ class InversionData(InversionLocations):
         self.components, self.data, self.uncertainties = self.get_data()
 
         self.locs = super().get_locs(self.params.data_object)
+        self.locs = super().z_from_topo(self.locs)
+
         self.mask = np.ones(len(self.locs), dtype=bool)
 
         if self.window is not None:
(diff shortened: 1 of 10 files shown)


geoapps/drivers/components/data.py
```diff
@@ -103,14 +102,15 @@ def __init__(self, workspace: Workspace, params: Params, window: Dict[str, Any])
...
-        """ Extract data from workspace using params data. """
+        """ Extract data from the workspace using params data. """
 
         self.vector = True if self.params.inversion_type == "mvi" else False
         self.n_blocks = 3 if self.params.inversion_type == "mvi" else 1
         self.ignore_value, self.ignore_type = self.parse_ignore_values()
         self.components, self.data, self.uncertainties = self.get_data()
 
         self.locations = super().get_locations(self.params.data_object)
+        self.locations = super().z_from_topo(self.locations)
```
domfournier, line 113: Does it not always drape data on topo?!? I don't think we want that...

domfournier: Also, `z_from_topo` is used as a Params/property in other classes. This could be confusing. Maybe rename to `drape_on_topo`?

domfournier (changes requested):
Maybe I don't get the code, but something is up I think with the initialization.

geoapps/drivers/components/data.py
benk-mira: Thank you for catching that - checking for z_from_topo now and calling a renamed set_z_from_topo method when True.

geoapps/drivers/components/locations.py
```diff
@@ -110,6 +111,12 @@ def get_locations(self, uid: UUID) -> np.ndarray:
             msg += " Object type should be Grid2D or point-like."
             raise (ValueError(msg))
 
+        if data_object.uid == self.params.topography_object:
+            if self.params.topography is not None:
+                elev = self.workspace.get_entity(self.params.topography)[0].values
+                if not np.all(locs[:, 2] == elev):
+                    locs[:, 2] = elev
+
```
domfournier, line 119: I am not following here. Why over-writing the locs Z on topo? Looks like a repeat from the initialize of the InversionTopography class

domfournier (approved):
Just check comment below. Looking good otherwise

geoapps/drivers/components/locations.py
benk-mira: Looks like I meant to remove one of them.  I opted to extend the get_locations method on the topography class to handle this, and removed from the locations.

domfournier (approved):
Approved!

PR #122, opened 2021-07-12
GEOPY-276: forward only option broken for some inversions

benk-mira:
**GEOPY-276 - forward only option broken for some inversions** 
 Forward only simulates data and writes results to workspace.  Simulation is very slow compared to the inversion.


domfournier:
Missing object type BlockModel
image

domfournier:
Data values not showing from Points entity
image

It does show for grid though
image

domfournier (changes requested):
Getting some pretty slow runs here too. I ll investigate a bit.

geoapps/drivers/components/data.py
```diff
@@ -343,3 +352,43 @@ def simulation(
...
+
+        sim, _ = self.simulation(mesh, active_cells)
+        d = sim.dpred(model)
+        d = d.compute()
+
+        if np.array(d).ndim == 1:
+            d = [d]
+
+        for i, c in enumerate(self.components):
+            self.predicted[c] = d[i]
```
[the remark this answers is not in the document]
benk-mira: Fixed up now and tested with tmi, bz channels.

geoapps/io/Gravity/constants.py
```diff
@@ -68,6 +69,13 @@
         ],
         "value": None,
     },
+    "gz_channel_bool": {
+        "default": False,
+        "group": "Data",
+        "main": True,
+        "label": "use Gz?",
```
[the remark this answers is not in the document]
benk-mira: Fixed

benk-mira (commented):
Running faster now

.gitignore
```diff
@@ -109,6 +109,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.conda-env
```
sebhmg, line 112: this is already excluded (see last line of the file)

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 
 default_language_version:
-    python: python3.7
+    python: python3
```
sebhmg, line 3: we want some control on the version here for reproducibility across developer machines. Forcing to 3.7 was good I think

But now, I see it want changes on geoh5py too from 3.7 to 3 ???

benk-mira: It was breaking with python3.7

.gitignore
sebhmg: the last commit on develop was removing that line...
no need to reintroduce it

.pre-commit-config.yaml
sebhmg: the last commit on develop was about forcing to 3.7
Please do not change.
We do not want pre-commit to auto-fix for a syntax which is reserved to 3.9 and not compatible with 3.7

.pre-commit-config.yaml
```diff
@@ -13,6 +13,7 @@ repos:
     rev: 21.7b0
     hooks:
     -   id: black
+        language_version: python3
```
sebhmg, line 16: not needed: it uses the `default_language_version` (supposed to be 3.7)

sebhmg (changes requested):
I am not reviewing the actual code but beware of unrelated change in .gitignore and pre-commit-config
Please see comments there. It seems changes in these two files have to be reverted

.pre-commit-config.yaml
benk-mira: When I revert to the python3.7 for both of your suggestions, I get a commit error:

"git" commit -F "C:\Users\Benjamin\Documents\Git\geoapps\.git\COMMITMESSAGE"
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('c:\\users\\benjamin\\anaconda3\\envs\\geoapps\\python.exe', '-mvirtualenv', 'C:\\Users\\Benjamin\\.cache\\pre-commit\\repotomysriv\\py_env-python3.7', '-p', 'python3.7')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'

stderr: (none)
Check the log at C:\Users\Benjamin\.cache\pre-commit\pre-commit.log
Done

Press Enter or Esc to exit...

benk-mira: https://github.com/pre-commit/pre-commit/issues/1375
^^^ this is what led us to try python3 instead of python3.7

sebhmg: maybe it requires some python 3.7 in the path?

well, I guess it is OK with Python 3 then. We will see...
I think the tool introducing syntax change is pyupgrade, and we tell this one to be compatible with 3.7

so keep global `default_language_version` as python3, but remove `language_version` here

benk-mira: Was able to commit with your recommended changes, but doesn't seem like it's triggering the tests still.

sebhmg (approved):

domfournier (approved):
Big changes. Still running. Marching on.

PR #127, opened 2021-07-14
GEOPY-288

sebhmg:
**GEOPY-288 - precommit in precommit-ci instead of github action**

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -17,33 +17,51 @@ on:
       - hotfix/**
 
 jobs:
-  pre-commit:
-    name: pre-commit
+  pylint:
+    name: pylint
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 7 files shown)


sebhmg:
@domfournier , @benk-mira, here I am adding pylint as a pre-commit hook.
It assumes the conda environment is under .conde-env
You can run `devtools/setup-dev.bat` to create it, and then change Python interpreter in pyCharm to use this conda env.
(this also has the benefit of having a different env fo each work tree, and it is not conflicting with your tests of a regular user installation)

pylint reports many errors or check failures, but still passes (for now) as long as score is above 5 (out of 10)

up for discussion. Your feedback is welcome!

domfournier (dismissed):
Seems to work on this side.

benk-mira (approved):
Tested by running setup-dev.bat, pointing pycharm to the new environment python.exe, running some code and pushing a small change that fired the linter.

PR #140, opened 2021-08-11
GEOPY-303

domfournier:
**GEOPY-303 - Inversion App: Topography data value crash**


benk-mira (approved):
Looks good to me!  Reproduced the error in another branch and tested in 303 by opening the pf_inversion notebook deselecting the topo channel field and running the inversion.

PR #141, opened 2021-08-11
GEOPY-275

domfournier:
**GEOPY-275 - data_type of observed data must match predicted data on save (to match colorscales in Analyst)**


benk-mira (approved):
Working for pf_inversion, but not MVIInversion and GravityInversion drivers.  Will need to address these in later issue

PR #142, opened 2021-08-12
GEOPY-273 - Add mesh from params option

benk-mira:
**GEOPY-273 - Add mesh from params option to base driver**


domfournier (approved):
Looking pretty slick

PR #144, opened 2021-08-16
GEOPY-244

domfournier:
**GEOPY-244 - Update write_ui_json to check for uuid strings and write to property field instead of value field**


geoapps/io/params.py
```diff
@@ -214,8 +214,11 @@ def to_dict(self, ui_json: dict = None, ui_json_format=True):
                 if isinstance(ui_json[k], dict):
                     field = "value"
                     if "isValue" in ui_json[k].keys():
-                        if ui_json[k]["isValue"] is False:
+                        if isinstance(new_val, UUID):
+                            ui_json[k]["isValue"] = False
```
benk-mira, line 218: Is this correct in the case of a parent object?  We, or possibly just I, have been using value to store the UUID of objects to which data 'channels' are added.  Eg: topography_object -> value: some_uuid, topography -> property: some_uuid, value: some_float.

domfournier: yea but the parent won't have `isValue`  in the dict, so we know that only applies for the channel field?

benk-mira: yep, you're right!

benk-mira (approved):
Thanks for the updates and tests!

benk-mira (approved):

PR #145, opened 2021-08-16
Geopy 284

benk-mira:
**GEOPY-284 - Forward Model as a separate driver** 
 General default ui cleanup and separation between forward and inverse ui.  Updated write_ui_json with a forward flag to isolate ui elements to those needed for foward modelling.  Forward and Inverse runs still share the same driver.


geoapps/io/input_file.py
```diff
@@ -166,14 +167,19 @@ def write_ui_json(
...
 
         out = deepcopy(ui_dict)
+        if forward:
+            out["forward_only"] = True
+            inversion_type = out["inversion_type"]
+            if inversion_type == "mvi":
+                from geoapps.io.MVI.constants import forward_parameters as fparams
+            if inversion_type == "gravity":
+                from geoapps.io.Gravity.constants import forward_parameters as fparams
+            out = {k: out[k] for k in fparams}
```
domfournier, line 182: Brilliant. So we are basically just sub-sampling the full UI right?

domfournier (approved):
That looks nice and easy.

geoapps/io/input_file.py
benk-mira: Exactly - doesn't fit the description of the issue, but I think it accomplishes what we needed with less code duplication than if we wrote a separate driver.

PR #148, opened 2021-08-23
Geopy 289

benk-mira:
**GEOPY-289 - Create mag sus inversion driver** 
 Abstracted common inversion parameters into an InversionParams class.  Subclassed the GravityParams/MVIParams and then added MagneticScalarParams (also renamed MVIParams -> MagneticVectorParams).


environment.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
   - ipykernel
   - scikit-image
   - scikit-learn
+  - pooch==1.4.0
```
domfournier, line 17: Looks like we might be in the clear. Remove and re-run tests.

geoapps/drivers/components/factories.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, params: Params):
         self.inversion_type = params.inversion_type
         from SimPEG import dask
 
-        if self.inversion_type == "mvi":
+        if self.inversion_type in ["mvi", "magnetic"]:
```
domfournier, line 49: Can we mirror the naming of the classes? Maybe flags: "magnetic vector", "magnetic scalar"?

domfournier: Will be handled on #GEOPY-290

PR #151, opened 2021-09-01
GEOPY-310

sebhmg:


sebhmg:
pre-commit.ci run

sebhmg:
see https://github.com/MiraGeoscience/geoapps/pull/152 on upstream instead

PR #155, opened 2021-09-10
GEOPY-282

domfournier:
**GEOPY-282 - Verify mvi solution and write test against accepted solution** 
 - Add run test for susceptibility forward and inverse
- Add run test for vector mag forward and inverse


tests/run_tests/driver_mag_test.py
```diff
@@ -0,0 +1,246 @@
...
+    target = {
+        "data_norm": 716.999,
+        "phi_m": 0.006225,
+        "phi_d": 92500.0,
+    }
+    # Run the forward
+    workspace = setup_workspace(tmp_path, 0.05)
+    model = workspace.get_entity("model")[0]
+    params = MagneticScalarParams(
+        forward=True,
```
benk-mira, line 97: Is this going to change since you updated the way we handle writing forward only drivers??

tests/run_tests/driver_mag_test.py
```diff
@@ -0,0 +1,246 @@
...
+    ##########################################################################
+    # Solution should satisfy this condition if run to completion.
+    # To get validated with full run if values in 'target' need to be updated.
+    #
+    # residual = (
+    #         np.linalg.norm(driver.inverse_problem.model - model) /
+    #         np.linalg.norm(model) * 100.
+    # )
+    # assert residual < 0.1, (
+    #     f"Deviation from the true solution is {residual}%. Please revise."
```
benk-mira, line 161: Nice touch!  I guess you can just run this a bunch of times in debug mode??  This doesn't look like it will run as is though, right?  driver doesn't contain a .inverse_problem attribute - is there a model in the output dictionary we can use in place of this?

tests/run_tests/driver_mag_test.py
```diff
@@ -0,0 +1,246 @@
...
+    b = 100
+    A = 50
+    zz = A * np.exp(-0.5 * ((xx / b) ** 2.0 + (yy / b) ** 2.0))
+    topo = np.c_[utils.mkvc(xx), utils.mkvc(yy), utils.mkvc(zz)]
+    triang = Delaunay(topo[:, :2])
+
+    Surface.create(workspace, vertices=topo, cells=triang.simplices, name="topography")
+
+    # Observation points
+    xr = np.linspace(-100.0, 100.0, 20)
```
benk-mira, line 35: I think we talked about a really extreme coarsening at one point.  Now that you've played around with the test, do you think we could get away with a single observation point and really coarse mesh?

benk-mira (dismissed):
Supa-cool!  Looks like this should be nice and reproducible for the other data types.  I'm curious to see if we can coarsen the solution a lot more since we're just interested in catching when the code changes for a given set of input.

tests/run_tests/driver_mag_test.py
benk-mira: we'll handle during merge with develop after 290 is in.

tests/run_tests/driver_mag_test.py
benk-mira: Actually it doesn't work for another reason - model here is a FloatData object and needs to be .value'ed, active_cell mapped, nan removed, and possibly reordered.

benk-mira (approved):
Looking good!

PR #157, opened 2021-09-15
Re-enable output of PF Inversion iteration convergence progress

nwilliams-kobold:
`SaveOutputEveryIteration` output directive was commented out of PF inversion code, likely to avoid a bug in the output formatting. This fixes the bug and enables the directive again.


domfournier (approved):
Looking good. Just keep in mind that both the simpegPF module and pf_inversion will be deprecated at some point.

domfournier:
Good on your side @nwilliams-kobold

PR #160, opened 2021-09-21
GEOPY-283

domfournier:
**GEOPY-283 - Verify grav solution and write test against accepted solution** 
 Unitest for gravity inversion.


benk-mira (approved):
A lot like the mag test one I just reviewed - approved!

PR #163, opened 2021-09-23
GEOPY-264

domfournier:
**GEOPY-264 - Unit test: Create 2D trend**


tests/utils_test.py
```diff
@@ -299,3 +301,31 @@ def test_filter_xy():
...
+    # Should return a plane even for order=5
+    comp_trend, comp_params = calculate_2D_trend(xy, nan_values, order=5, method="all")
+    np.testing.assert_almost_equal(values, comp_trend)
+    # Should return same plane parameter for 'corners' or 'all'
+    corner_trend, corner_params = calculate_2D_trend(
+        xy, nan_values, order=1, method="corners"
+    )
+    np.testing.assert_almost_equal(values, corner_trend)
+
+    with pytest.raises(ValueError):
```
benk-mira: We can test the specific exception handling here with a small modification:
`with pytest.raises(ValueError) as excinfo:
    calculate_2D_trend(...)
 assert "part of your custom raise message here" in str(excinfo.value)
`

benk-mira: I've pushed the edits for this one.

benk-mira (approved):
pushed a small change to make the test a little stronger, but looks great otherwise.

PR #167, opened 2021-09-29
GEOPY-268

benk-mira:
**GEOPY-268 - warning about KMean OMP and memory leak ** 
 kmeans warning suppression is local to the clustering call and specific to the category of warning issued.


domfournier (approved):
Easy fix, great!

PR #168, opened 2021-09-29
GEOPY-129

benk-mira:
First pass at writing test that writes blockModel to tensor, then back again to compare entities.  Test fails for negative z, or with rotation.

### DIFF
--- a/tests/utils_test.py
+++ b/tests/utils_test.py
@@ -11,14 +11,18 @@
 #  (see LICENSE file at the root of this source code package).
 
 import itertools
+import os
 
 import numpy as np
 import pytest
 from discretize import TreeMesh
+from geoh5py.objects import BlockModel
+from geoh5py.shared.utils import compare_entities
 from geoh5py.workspace import Workspace
 
 from geoapps.utils.testing import Geoh5Tester
 from geoapps.utils.utils import (
+    block_model_2_tensor,
     calculate_2D_trend,
     downsample_grid,
     downsample_xy,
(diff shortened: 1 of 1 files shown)


tests/utils_test.py
```diff
@@ -332,3 +337,132 @@ def test_detrend_xy():
...
+    x = np.linspace(0, 10, nx)
+    y = np.linspace(0, 10, ny)
+    z = np.linspace(0, 10, nz)
+
+    blockmodel_tensormesh_equivalency(ws, [0.0, 0.0, 0.0], x, y, z)
+    blockmodel_tensormesh_equivalency(ws, [10.0, 10.0, 10], x, y, z)
+    blockmodel_tensormesh_equivalency(ws, [-10.0, -10.0, -10], x, y, z)
+
+
+def test_block_model_2_tensor_negativeh(tmp_path):
```
domfournier, line 430: very canadian of you ... but can we fix the name eh?

tests/utils_test.py
```diff
@@ -332,3 +337,132 @@ def test_detrend_xy():
...
+            "DataValues": {
+                "association": "CELL",
+                "values": np.random.randint(
+                    0, 100, size=len(block_model.centroids)
+                ).astype(np.int32),
+            }
+        }
+    )
+
+    tensor_mesh, model = block_model_2_tensor(block_model, models=[data.values])
```
domfournier, line 464: Were we going to test the model array?

tests/utils_test.py
benk-mira: haha, the name is negative h as in cell size.  changed to negative_z for clarity

domfournier (approved):
image

PR #169, opened 2021-10-01
GEOPY-338

benk-mira:
**GEOPY-338 - Forward modeling of Vector model broken** 
 Fix erroneously resorted model and update the test to look at a heterogeneous model instead of constant.

### DIFF
--- a/geoapps/drivers/components/models.py
+++ b/geoapps/drivers/components/models.py
@@ -370,7 +370,7 @@ class InversionModel:
         else:
             xyz_in = parent.vertices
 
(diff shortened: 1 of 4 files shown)


geoapps/drivers/components/models.py
```diff
@@ -370,7 +370,7 @@ def _obj_2_mesh(self, obj, parent):
         else:
             xyz_in = parent.vertices
 
-        xyz_out = self.mesh.original_cc()
```
domfournier, line 373: Are we going to ever use original_cc?

domfournier (approved):

PR #170, opened 2021-10-04
GEOPY-271

benk-mira:
**GEOPY-271 - Inversions: default output air cells to np.nan ** 
 Added assert to run test to keep tests lean.


domfournier (approved):
all nans all the time!

PR #172, opened 2021-10-08
GEOPY-304

domfournier:


geoapps/io/DirectCurrent/constants.py
```diff
@@ -53,10 +53,10 @@
     "max_distance": np.inf,
     "horizontal_padding": 1000.0,
     "vertical_padding": 1000.0,
-    "window_center_x": 0.0,
-    "window_center_y": 0.0,
-    "window_width": 0.0,
-    "window_height": 0.0,
+    "window_center_x": None,
```
benk-mira, line 56: These need to be 0.0 or else we get strings back from ui.json

benk-mira (changes requested):
Just one comment regarding the window values being None.  This will cause the values to be read in as strings when running from GA.

benk-mira:
Doesn't need to be a part of this issue, but I get a deprecation warning for Simped @ SimPEG/utils/code_utils.py:413, 420 for Solver (please use simulation.solver)

geoapps/io/DirectCurrent/constants.py
domfournier: Done

PR #176, opened 2021-10-19
GEOPY-351

sebhmg:


domfournier:
Might have missed one. But the data transfer app has been around for a year...

domfournier:
DCIP pushed now.

Install_or_Update.bat
```diff
@@ -1,8 +1,10 @@
-set PATH=%PATH%;%USERPROFILE%\anaconda3\Scripts;
-call activate.bat
-call conda remove --name geoapps --all
-call conda env update --file environment.yml  --prune
-call activate.bat geoapps
-call conda env config vars set KMP_WARNINGS=0
+set CONDA_EXE=conda
+:: if Conda executable is not in the PATH, uncomment and set the executable location below
+::set CONDA_EXE=%USERPROFILE%\AppData\Local\Continuum\anaconda3\Library\bin\conda.bat
+
```
domfournier, line 4: Looks like you need to call the activate.bat. I m getting this error otherwise :
image

domfournier: add `call %CONDA_EXE% activate`

Install_or_Update.bat
```diff
@@ -1,8 +1,10 @@
...
+call %CONDA_EXE% remove --name geoapps --all --yes
+call %CONDA_EXE% env update --file environment.yml --prune
+call %CONDA_EXE% env config vars set KMP_WARNINGS=0 --name geoapps
+call %CONDA_EXE% activate geoapps && python -m pip install -e .
```
domfournier, line 8: Got an issue here with this line. The pip install of geoapps tries to re-install h5py from pip, which seems to crash on wheels: 
image

domfournier: Just need to change to this:
`pip install -e . --no-deps`

sebhmg: strange, I did not need that.
pip normally detects the dependency is already there

sebhmg: anyway, dependencies are installed already in the env, so `--no-deps` makes sense. Thanks !

Install_or_Update.bat
sebhmg: not sure what happened in you env, but this is supposed to call conda.bat, never conda.exe
image

Might be fixed with the script I have introduced to find Conda.
Passing the env name to conda conf with `-n` or `--name` should work: `conda env config vars set my_var=value -n test-env` (from [Conda doc](
https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables))

domfournier: ok sounds good, I ll test

sebhmg: also I have moved this variable definition directly in the environment .yml file, so this command line is not necessary anymore

Install_or_Update.bat
```diff
@@ -1,8 +1,14 @@
...
+call %~dp0get_conda_exec.bat
+if !errorlevel! neq 0 (
+  exit /B !errorlevel!
+)
+
+cd %~dp0
+call !MY_CONDA_EXE! remove --name geoapps --all --yes
+call !MY_CONDA_EXE! env update --file environment.yml --prune
+call !MY_CONDA_EXE! activate geoapps && python -m pip install -e . --no-deps
 pause
```
domfournier, line 13: Ok not sure what I am doing wrong here, but I am just getting this when running the bat file
image

No installation triggered?

sebhmg: but it did not fail, you would not have the "Using Conda" message

domfournier: yea I dont know man. Still not working here.

PR #183, opened 2021-10-26
GEOPY-358: followup: fix running pylint from precommit

sebhmg:
**GEOPY-358 - use local editable geoh5 in geoapps dev environment**


benk-mira (approved):
Looks fine to me.  Will be tested on fresh windows install.

domfournier (approved):
if you say so

PR #190, opened 2021-11-03
GEOPY-319-b

domfournier:
**GEOPY-319 - Verify IP solution and write test against accepted solution**


geoapps/drivers/base_inversion.py
```diff
@@ -379,6 +381,9 @@ def get_tile_misfits(self, tiles):
             lsim, lmap = self.inversion_data.simulation(
                 self.mesh, self.active_cells, lsurvey, tile_id
             )
+
+            # TODO Parse workers to simulations
```
benk-mira, line 385: Is this still a TODO?  aren't you assigning workers to the simulation in the next line?

geoapps/io/InducedPolarization/constants.py
```diff
@@ -218,6 +220,8 @@
             if default_ui_json[k]["isValue"] == False:
                 key = "property"
         default_ui_json[k][key] = v
+        if "enabled" in default_ui_json[k].keys() and v is not None:
```
benk-mira, line 223: I don't think this will work with geopy-318.  There will be times where we want the value to be not None with enabled false.  See the none_map feature in 318.

benk-mira (commented):
There's just one hangup in the induced polariation constants.py that won't jive with GEOPY-318.  Let me know if you think the line needs to be in for ip to work (it's not in any of the other constants.py files).

geoapps/drivers/base_inversion.py
domfournier: Yes, will need to assign a specific worker to simulations. Right now it's just None anyway

geoapps/io/InducedPolarization/constants.py
domfournier: Done. Just need to wait for the tests to pass

benk-mira (dismissed):
Superb

PR #197, opened 2021-11-04
GEOPY-358: fix script to start app in dev env [WIP]

sebhmg:
**GEOPY-358 - use local editable geoh5py in geoapps dev environment**


devtools/start-app.bat
```diff
@@ -7,6 +7,7 @@ if !errorlevel! neq 0 (
   exit /B !errorlevel!
 )
 
+set PYTHONPATH=%project_dir%;%PYTHONPATH%
```
sebhmg, line 10: from PyCharm terminal, `import geoapps` is working because `geoapps` is in the current folder...

andrewg-mira (approved):

devtools/start-app.bat
benk-mira: Thanks Seb, that last one did the trick.  Tested by printing a line in geoh5 init then running devtools/start-app, run first cell of inversion app and getting a print statement (nothing added to environment variables).

benk-mira (approved):
Works as expected now.  Thanks for the start-app script, too!

sebhmg:
> Works as expected now. Thanks for the start-app script, too!

@benk-mira , thanks for the test.
I have pushed yet another change (installing geoapps editble in the local conda env, same as done for a user installation).
I think this way is better as you will also be able to have it working from PyCharm terminal, typing there: `jupyter notebook geoapps\applications\Index.ipynb`

benk-mira:
image

benk-mira (changes requested):

sebhmg:
will create a new PR

PR #204, opened 2021-11-08
Release/0.6.0

domfournier:


sebhmg (changes requested):
as discussed, will need to lock requriments on a tag for "simpeg archive", and on a version branch for "simpeg tiled simulations"

sebhmg (approved):
🚀

PR #207, opened 2021-11-09
Geopy 344

benk-mira:
**GEOPY-344 - params write_ui_json fails if params class constructed without kwargs** 
 Discovered that the blank constructor was working.  Beefed up params tests that uses blank constructed classes to check that defaults can be written to file, read back into an input file and produce a params class through both the from_path and from_input_file options that contain the default values.


domfournier (approved):
Looking good here.

PR #216, opened 2021-11-16
GEOPY-394

benk-mira:
**GEOPY-394 - Not all window params should be optional, just groupOptional** 
 Handle None values in the none_map differently for groupOptional groups.  Involves finding out if group is optional - introduced a new class for uijson utils and executed here.  When group is found to be optional and in the none_map, parameters are made enabled false without optional true.


geoapps/io/input_file.py
```diff
@@ -401,3 +391,104 @@ def _set_associations(self, d: dict[str, Any]) -> None:
...
+                else:
+                    data[k] = v[field]
+            else:
+                data[k] = v
+
+        return data
+
+    def data(self):
+        """Applies _data to self.ui."""
+        return UIJson._data(self.ui)
```
domfournier, line 420: Should we not stash the data once we have done it once?

geoapps/io/input_file.py
```diff
@@ -401,3 +391,104 @@ def _set_associations(self, d: dict[str, Any]) -> None:
...
+                    if value is None:
+                        data[k] = v
+                    else:
+                        if v[field] == value:
+                            data[k] = v
+        return data if data else None
+
+    def collect(self, field: str, value: Any = None) -> dict[str, Any]:
+        """Applies _collect to self.ui."""
+        return UIJson._collect(self.ui, field, value)
```
domfournier, line 438: same here, looks like we are going to loop again everytime we ask for the `collect`

domfournier (commented):
Looks like a step forward for sure. See comments below.

geoapps/io/input_file.py
benk-mira: I think you're right.. I'll just have to think through what happens to the dictionary outside the scope of UIJson.ui if I edit it inside.  Maybe I should do a deepcopy on init?

geoapps/io/input_file.py
benk-mira: In this case, I think this would be the correct behaviour.  This method just collects and returns groups of parameters.  What you're describing would be like a 'filter' method - could also be useful

geoapps/io/input_file.py
benk-mira: no wait... that would destroy the uijson format.  This method converts uijson format to simple key/value.

geoapps/io/input_file.py
domfournier: gotcha

geoapps/io/input_file.py
domfournier: maybe you just need to talk me through it

RichardScottOZ: You have a  couple of different if...else styles here too that is perhaps making this harder to parse for people?

```python
for k, v in d.items():
+            if isinstance(v, dict):
+                field = "value" if UIJson._truth(d, k, "isValue") else "property"   <----
+                if not UIJson._truth(d, k, "enabled"):
+                    data[k] = None
+                else:
+                    data[k] = v[field]
+            else:
+                data[k] = v
+
```

domfournier (approved):
Thanks for refactoring, looks great.

PR #220, opened 2021-11-24
Geopy 396

benk-mira:
**GEOPY-396 - use_ component ui elements snuck back into inversion ui.json files.** 
 Redesign of the constructors and .constants files to allow independent forward and inverse defaults and default_ui_json.


geoapps/base.py
```diff
@@ -139,8 +139,8 @@ def file_browser_change(self, _):
             _, extension = path.splitext(self.file_browser.selected)
 
             if extension == ".json" and getattr(self, "_param_class", None) is not None:
-                self.params = getattr(self, "_param_class").from_path(
-                    self.file_browser.selected
+                self.params = getattr(self, "_param_class")(
+                    InputFile(self.file_browser.selected)
```
domfournier, line 143: noice

geoapps/io/MagneticVector/constants.py
```diff
@@ -781,28 +804,30 @@
         "types": [str, UUID, int, float],
     },
     "reference_inclination_object": {
-        "types": [str],
+        "types": [str, UUID],
```
domfournier, line 807: Should we add FloatData right away like above?

geoapps/io/validators.py
```diff
@@ -385,7 +384,7 @@ def validate_input(self, input) -> None:
                         + f"{self.free_params_keys}"
                     )
 
-            input._free_params_dict = free_params_dict
+            # return free_params_dict
```
domfournier, line 387: cleanup

tests/input_validator_test.py
```diff
@@ -88,30 +88,30 @@ def test_validate_parameter_shape():
...
+#         True,
+#         "u_cell_size",
+#     )
+#     vtype = "reqs"
+#     input_keys = ["mesh_from_param", "topography"]
+#     with pytest.raises(KeyError) as excinfo:
+#         validator._validate_parameter_req(param, value, validations)
+#     msg = f"Unsatisfied '{param}' requirement. Input file must contain "
+#     msg += f"'{validations[1]}' if '{param}' is '{str(value)}'."
+#     assert msg in str(excinfo.value)
```
domfournier, line 114: cleanup ?

domfournier (approved):
Another great refresh. Minor edits posted below

geoapps/io/MagneticVector/constants.py
benk-mira: This param is just the object, so it should be just str, UUID.  The reference_inclination param would does already have the int, float types included.

domfournier: No I meant as an entity type

benk-mira: Ok, gotcha now.  This isn't required at the moment because the write_input_file method demotes geoh5py objects.  We can add these types to the validators down the road though.

geoapps/io/validators.py
benk-mira: Done

tests/input_validator_test.py
benk-mira: Added the test back in.

domfournier (approved):
Rolling...

PR #221, opened 2021-11-29
Geopy 391

benk-mira:
Added test to check that no references to default field are left in the .constants files.


domfournier (approved):
looking good

domfournier (approved):

PR #222, opened 2021-11-30
Geopy 392

benk-mira:
Visible field is now removed from geoapps since GA doesn't update this field and not visible params can be accomplished with simple key/value pairs


domfournier (approved):
Fine by me

domfournier (approved):

PR #226, opened 2021-12-01
GEOPY-396

domfournier:
**GEOPY-396 - use_ component ui elements snuck back into inversion ui.json files.**


benk-mira (approved):
whoops!

PR #227, opened 2021-12-02
GEOPY-384: Error re-loading gravity inversion

benk-mira:
**GEOPY-384 - Error re-loading gravity inversion**

Updated write_default_uijson with option to write files defaulted with app_initializer values because I am tired of the same values into a ui all the time.  More importantly, I used the new mechanism to write a ui.json and run from GA so that I could test the bug on this issue by loading the ui.json into the notebook and re-running there.  Doesn't seem to be an issue now that GEOPY-396 is resolved


domfournier (approved):
Looking good.

domfournier (approved):

PR #228, opened 2021-12-06
Geopy 387

benk-mira:
**GEOPY-387 - Remove duplicated workspaces: geoh5 and workspace attribute.** 
 Removing workspace parameter name in favor of geoh5.  Apps and inversion drivers will still refer to workspace though


domfournier (approved):
Thanks for this, much nicer

PR #230, opened 2021-12-06
Geopy 405

benk-mira:
**GEOPY-405 - Data transfer applies padding to south extent when north is provided** 
 Fix the padding behaviour vs labelling of padding options.  Switched north and south labels


domfournier (approved):
Easy peasy

PR #231, opened 2021-12-06
Geopy 395

benk-mira:
**GEOPY-395 - save_misfit is not working for multi-component inversions** 
 Removed finish_inversion_message and collect_predicted_data that was generating the error reported in the issue.


domfournier (approved):
Great.

PR #233, opened 2021-12-07
Geopy 407

sebhmg:
so let's see if this comment passes! Seems a good test with something "special" characters *@ and {{more}}


sebhmg:
trying another PR

PR #248, opened 2021-12-09
move write_default_uijson out of import in init to remove runtimewarning

benk-mira:
hotfix to get rid of a warning encountered during demo practise


domfournier (approved):
cool. Not sure I get why it failed, but fine by me.

PR #252, opened 2021-12-10
GEOPY-390

benk-mira:
**GEOPY-390 - Add a write_default_ui_json method in root of io module that finds all sub-modules and writes the a defaulted ui.json file for each.**
Requested fix to include peakfinder and octree files when run with use_initializers option

### DIFF
--- a/geoapps/io/write_default_uijson.py
+++ b/geoapps/io/write_default_uijson.py
@@ -86,7 +86,9 @@ def write_default_uijson(path, use_initializers=False):
     for filename, params in filedict.items():
 
         if use_initializers:
-            if getattr(params, "forward_only", None) is not None:
+            if type(params) in [OctreeParams, PeakFinderParams]:
+                params.write_input_file(name=filename, path=path)
+            elif getattr(params, "forward_only", None) is not None:
                 params.write_input_file(name=filename, path=path)
         else:
             params.write_input_file(name=filename, path=path, default=True)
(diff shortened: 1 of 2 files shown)


geoapps/io/write_default_uijson.py
```diff
@@ -86,7 +86,9 @@ def write_default_uijson(path, use_initializers=False):
     for filename, params in filedict.items():
 
         if use_initializers:
-            if getattr(params, "forward_only", None) is not None:
+            if type(params) in [OctreeParams, PeakFinderParams]:
+                params.write_input_file(name=filename, path=path)
+            elif getattr(params, "forward_only", None) is not None:
```
domfournier, line 91: I m still not getting the logic. When are we not writing with initializer?

domfournier: Can we just set `default = not use_initializers`

geoapps/io/write_default_uijson.py
```diff
@@ -84,12 +84,8 @@ def write_default_uijson(path, use_initializers=False):
     }
 
     for filename, params in filedict.items():
-
-        if use_initializers:
-            if getattr(params, "forward_only", None) is not None:
-                params.write_input_file(name=filename, path=path)
-        else:
-            params.write_input_file(name=filename, path=path, default=True)
+        default = False if use_initializers else True
```
domfournier, line 87: Not to be picky, but that won't be mypy compliant. Convert to `not use_initializer` for simpler logical

domfournier (approved):

geoapps/io/write_default_uijson.py
benk-mira: Ternary's aren't mypy compliant?

domfournier: either pylint or mypy, can't remember which but I had to replace every where on geoh5py.

PR #253, opened 2021-12-14
Geopy 404

benk-mira:
**GEOPY-404 - Switching between interpolation algorithms in data transfer app causes a dimension mismatch error**
Began issue by reorganizing the trigger_click function to better understand what was happening.  Created a few small staticmethods and added tests to confirm the expected behaviour.  

During the testing I discovered that adding padding to the up direction shifted the core region down.  I fixed this by shifting the origin by the padding amount for this case.

After the reorganization could not reproduce the shape mismatch error.

### DIFF
--- a/geoapps/utils/utils.py
+++ b/geoapps/utils/utils.py
@@ -43,7 +43,10 @@ from sklearn.neighbors import KernelDensity
 
 def get_locations(workspace: Workspace, object: UUID | Entity):
     """
-    Returns locations of data object centroids or vertices.
+    Returns object's centroids or vertices.
+
+    If no location data is found on the provided object, the method will
+    attempt to call itself on it's parent.
 
     :param workspace: Geoh5py Workspace object.
     :param object: Object or uuid of object containing centroid or
@@ -61,7 +64,10 @@ def get_locations(workspace: Workspace, object: UUID | Entity):
     elif hasattr(object, "vertices"):
         locs = object.vertices
     else:
-        locs = None
+        if object.parent is not None:
+            locs = get_locations(workspace, object.parent)
+        else:
+            locs = None
 
     return locs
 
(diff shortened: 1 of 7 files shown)


geoapps/utils/utils.py
```diff
@@ -40,6 +41,37 @@
...
+    :param workspace: Geoh5py Workspace object.
+    :param object: Object or uuid of object containing centroid or
+        vertex location data.
+
+    :return: Array shape(*, 3) of x, y, z location data
+
+    """
+
+    if isinstance(object, UUID):
+        object = workspace.get_entity(object)[0]
```
domfournier, line 60: This call could return None if the UUID isn't in the workspace, which will make Line 67 to crash without proper error. See suggestion below.

geoapps/utils/utils.py
```diff
@@ -40,6 +41,37 @@
...
+
```
domfournier, line 61:
```suggestion
locs = None
```

geoapps/utils/utils.py
```diff
@@ -40,6 +41,37 @@
...
+    if hasattr(object, "centroids"):
+        locs = object.centroids
+    elif hasattr(object, "vertices"):
+        locs = object.vertices
+    else:
```
domfournier, line 66:
```suggestion
    elif getattr(object, "parent", None) is not None:
```

geoapps/utils/utils.py
```diff
@@ -40,6 +41,37 @@
...
+        if object.parent is not None:
+            locs = get_locations(workspace, object.parent)
+        else:
```
domfournier, line 69:
```suggestion

```

geoapps/utils/utils.py
```diff
@@ -40,6 +41,37 @@
...
+            locs = None
```
domfournier, line 70:
```suggestion

```

domfournier (commented):
This is epic. Thanks for all the unit tests. Minor edit suggested below.

PR #254, opened 2021-12-14
Geopy 414

benk-mira:
**GEOPY-414 - Fix sorting of iterations in data transfer app**
Wrote a util to fix the order children with iteration data and dumped data into the options list of ObjectDataSelection


geoapps/utils/utils.py
```diff
@@ -41,6 +41,38 @@
...
+    iteration_data = []
+    other_data = []
+    for c in object.children:
+        if not isinstance(c, (IntegerData, FloatData)):
+            continue
+        else:
+            children_dict[c.name] = c.uid
+            name = c.name
+            syllables = name.split("_")
+            if any([n.lower() == "iteration" for n in syllables]):
```
domfournier, line 60: This is a bit too target to inversion. Consider using a regexp to detect any float or integer. Something like 
`re.findall(""[-+]?[0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?", name)` should return all decimal numbers including exponential format from the string `name`.

benk-mira: fair enough, done.

geoapps/utils/utils.py
```diff
@@ -41,6 +41,37 @@
...
+        if not isinstance(c, (IntegerData, FloatData)):
+            continue
+        else:
+            name = c.name
+            children_dict[name] = c.uid
+            iter = re.findall(r"[-+]?[0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?", name)
+            if iter:
+                try:
+                    iters.append(int(iter[0]))
+                except:
```
domfournier, line 63: Surprised that mypy didn't complain about using bare except. Is there a reason we would need to operate on integer only? I would suggest always using float and removing the try/except if not necessary.

geoapps/utils/utils.py
```diff
@@ -41,6 +41,37 @@
...
+                    iters.append(float(iter[0]))
+                iteration_data.append(name)
+            else:
+                other_data.append(name)
+
+    iteration_data = np.array(iteration_data)[np.argsort(iters)].tolist()
```
domfournier, line 69: What would happen if the object has two sets of names with numerical value in it? Let's say we have Obs[1], Obs[2], ... and Pred[1], Pred[2] ... Looks like we are going to mix them both together the way it's done now? Feels like we need to keep track of the unique base names...

tests/utils_test.py
```diff
@@ -40,6 +41,44 @@
...
+    test_data = grid.add_data({"Iteration_1_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_5_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_3_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_2_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_4_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_9.0_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_8e0_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_11_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_6_data": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"Iteration_7_data": {"values": np.ones(10 * 15)}})
```
domfournier, line 68: Try augmenting the test with two sets. And we probably don't need that many.

tests/utils_test.py
```diff
@@ -40,6 +41,44 @@
...
+    test_data = grid.add_data({"topo": {"values": np.ones(10 * 15)}})
+    test_data = grid.add_data({"uncert": {"values": np.ones(10 * 15)}})
+
+    d = sorted_children_dict(grid)
+    d = list(d.keys())
+    assert d[0] == "Iteration_1_data"
+    assert d[1] == "Iteration_2_data"
+    assert d[7] == "Iteration_8e0_data"
+    assert d[8] == "Iteration_9.0_data"
```
domfournier, line 77: We are not testing the more critical case of 11 after 2

geoapps/utils/utils.py
benk-mira: removed

geoapps/utils/utils.py
benk-mira: handled

tests/utils_test.py
benk-mira: yep

tests/utils_test.py
benk-mira: am now

benk-mira:
Added a test that should have covered all the comment previously given

geoapps/utils/utils.py
```diff
@@ -41,6 +41,53 @@
...
+    try:
+        text_as_float = float(text)
+        text_as_int = int(text_as_float)
+        return text_as_int if text_as_int == text_as_float else text_as_float
+    except ValueError:
+        return text
+
+
+def sorted_alphanumeric_list(l: list[str]) -> list[str]:
+    """
```
domfournier, line 55: Single letter for variables isn't pep8 compliant.

geoapps/utils/utils.py
```diff
@@ -41,6 +41,53 @@
...
+    Sorts a list of string containing alphanumeric characters in readable way.
+
+    Sorting precedence is alphabetical for all string components followed by
+    numeric component found in string from left to right.
```
domfournier, line 59: should have a :param value : and `:returns:`  in docstring

geoapps/utils/utils.py
```diff
@@ -41,6 +41,53 @@
...
+    """
+
+    def sort_precedence(text):
+        numeric_regex = r"[-+]?[0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?"
+        non_numeric = re.split(numeric_regex, text)
+        numeric = [string_2_numeric(k) for k in re.findall(numeric_regex, text)]
+        order = non_numeric + numeric
```
domfournier, line 66: Oh very nice.

geoapps/utils/utils.py
benk-mira: ok

geoapps/utils/utils.py
benk-mira: yep

domfournier (approved):
A+

PR #256, opened 2021-12-15
Geopy 416

benk-mira:
**GEOPY-416 - Fix PeakFinder and Octree apps to add default group template params to default_ui_json definition.  **
**GEOPY-416 - Fix PeakFinder and Octree apps to add default group template params to default_ui_json definition. **
Added refinement a, b in default_ui_json. Removed Template from both peakfinder and octreemesh. Updated code to handle this new structure. Added to the Peakfinder app_initializer to be able to reproduce the notebook app default run. Added to the octree creation test to make sure that the new structure still allows to add refinement beyond the two in the default_ui_json


domfournier (approved):
Looking good. Ready for testing

PR #260, opened 2021-12-17
Geopy 423

benk-mira:
**GEOPY-423 - Octree mesh app fails to execute if geoh5 switched from default FlinFlon file.**
trigger_click() event was looping over widget values to set them into params and run the app.  For some reason the loop didn't reach the widget containing the workspace.  Params was stuck with the original workspace set during init.  Fix was to set params.geoh5 on the apps workspace.setter.


domfournier (approved):
Easy

PR #265, opened 2021-12-20
GEOPY-419

domfournier:
**GEOPY-419 - Inversion: Residual and observed label flipped**


benk-mira (changes requested):
I feel like you forgot to push some of your changes...  I quickly pulled 419 and tested and seem to still be broken.

benk-mira (approved):

PR #290, opened 2022-02-17
GEOPY-89

domfournier:
**GEOPY-89 - Implement MT inversion in geoapps**


geoapps/drivers/base_inversion.py
```diff
@@ -341,34 +342,35 @@ def get_tile_misfits(self, tiles):
             [],
         )
         for tile_id, local_index in enumerate(tiles):
```
benk-mira, line 344: Whoops, get_tile_misfits should have been replaced by MisfitFactory.  This should have been deleted

benk-mira (changes requested):
A quick scan

geoapps/drivers/base_inversion.py
domfournier: True, and deleted

benk-mira (approved):
all good

PR #292, opened 2022-03-02
GEOPY-445

domfournier:
**GEOPY-445 - Air cells not removed on starting model for DC inversion**


geoapps/drivers/components/models.py
```diff
@@ -356,7 +356,7 @@ def edit_ndv_model(self, model):
                 values[~model] = np.nan
                 data_obj[0].values = values
 
-        self.workspace.finalize
+        self.workspace.finalize()
```
benk-mira, line 359: haha, too simple!

PR #301, opened 2022-04-11
GEOPY-388

benk-mira:
**GEOPY-388 - Add a validation type to make sure at least one channel of data and uncertainty are provided to inversion.**


geoapps/inversion/potential_fields/gravity/constants.py
```diff
@@ -547,6 +547,26 @@
         "values": ["gravity"],
     },
     "data_object": {"types": [str, UUID, Points, Surface, Grid2D]},
+    "gz_channel": {"one_of": "data channel"},
+    "gz_uncertainty": {"one_of": "uncertainty channel"},
```
domfournier, line 551: Don't think we need the one_of on uncerts...

domfournier (commented):
I think we will need to do the same to mag_scalar, mvi, and MT?

domfournier (dismissed):
Looks good, I ll test again

geoapps/inversion/driver.py
```diff
@@ -368,17 +368,23 @@ def start_inversion(filepath=None, **kwargs):
 
     if inversion_type == "magnetic vector":
         from .potential_fields import MagneticVectorParams
+        from .potential_fields.magnetic_vector.constants import validations
 
+        input_file = InputFile.read_ui_json(filepath, validations=validations)
```
domfournier, line 373: Move this down after the logicals and wait to instantiate the Params.

benk-mira: actually can't move them down since we're capturing a different validations at the if block scope into a specific params.

domfournier: still all name validations though?

PR #303, opened 2022-04-11
GEOPY-481: fixup - protect against space in path

sebhmg:
**GEOPY-481 - Conflict with geosoft as default anaconda**

### DIFF
--- a/Start_applications.bat
+++ b/Start_applications.bat
@@ -6,5 +6,6 @@ if !errorlevel! neq 0 (
   exit /B !errorlevel!
 )
 
-call !MY_CONDA_EXE! activate geoapps && jupyter notebook geoapps\index.ipynb
+set MY_CONDA=!MY_CONDA_EXE:"=!
+call "!MY_CONDA!" activate geoapps && jupyter notebook geoapps\index.ipynb
 cmd /k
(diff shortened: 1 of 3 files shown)


Start_applications.bat
```diff
@@ -1,10 +1,11 @@
...
 setlocal EnableDelayedExpansion
 
-call %~dp0get_conda_exec.bat
+call "%~dp0get_conda_exec.bat"
 if !errorlevel! neq 0 (
   exit /B !errorlevel!
 )
 
-call !MY_CONDA_EXE! activate geoapps && jupyter notebook geoapps\index.ipynb
+set MY_CONDA=!MY_CONDA_EXE:"=!
```
andrewg-mira, line 9: Is this line missing a matching quote " character?

domfournier (approved):
Looking good. I will test it.

Start_applications.bat
domfournier: Loooks like it, nice catch

sebhmg: no, that is the syntax to remove existing quote in the path... before I wrap again the variable in quote when calling conda
(Do not ask my why, but conda fails if I just leave the quotes as is in the variable)

PS: did I mentioned already that I hate batch script?

andrewg-mira: haha yes I actually wasn't sure. It definitely looks weird.

andrewg-mira (approved):

PR #309, opened 2022-04-19
GEOPY-502

benk-mira:
**GEOPY-502 - Cannot run peak finder twice as it gives association validation error.**
Proposed fix is to add line_field from new_workspace to params_dict

### DIFF
--- a/geoapps/peak_finder/application.py
+++ b/geoapps/peak_finder/application.py
@@ -1289,7 +1289,7 @@ class PeakFinder(ObjectDataSelection):
         )
         for key, value in param_dict.items():
             if isinstance(value, ObjectBase):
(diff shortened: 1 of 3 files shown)


geoapps/peak_finder/application.py
```diff
@@ -1289,7 +1289,15 @@ def trigger_click(self, _):
         )
         for key, value in param_dict.items():
             if isinstance(value, ObjectBase):
-                param_dict[key] = value.copy(parent=new_workspace, copy_children=True)
+                if new_workspace.get_entity(value.uid) is not None:
```
domfournier, line 1292: Should it not be the opposite? If the object is None then we copy?

domfournier (approved):

PR #311, opened 2022-04-20
GEOPY-495

benk-mira:
**GEOPY-495 - Write simpeg output file like UBC**
Adds log and output file creation.  Relies on PR with same name on simpeg.


domfournier (approved):
All clear

domfournier (approved):

PR #314, opened 2022-04-22
GEOPY-92

benk-mira:
Adding an app_initializer to the MT .constants to reproduce the MT inversion I ran overnight last night.


benk-mira:
Re-opening with new GEOPY-92 branch

PR #317, opened 2022-04-25
Geopy 510

domfournier:
**GEOPY-510 - Fix/standardize app_initializers **


geoapps/octree_creation/application.py
```diff
@@ -27,7 +28,6 @@ class OctreeMesh(ObjectDataSelection):
     Widget used for the creation of an octree mesh
     """
 
-    defaults = {}
```
benk-mira, line 30: Is this the line that the issue is supposed to address?  That would make sense to me if defaults were imported above and the {} wiped them out, but that doesn't seem to be the case

benk-mira (commented):
This looks good to me.  But I'm not sure how these changes relate to the description in the issue:
"""
Initilization of triangulated surface application broken due to re-set of defaults (mutable property {}).

Need to standardize the app_initalizer and protect the defaults.
"""

geoapps/octree_creation/application.py
domfournier: Importing OctreeMesh would overwrite the base class defaults.

PR #331, opened 2022-05-09
GEOPY-521

jamiefb:
**GEOPY-521 - Create ui.json for iso_surface**


domfournier (approved):
Great! This will be done by the write_default_ui_json function eventually, but works for now.

PR #334, opened 2022-05-10
Geopy 513 standalone

domfournier:
**GEOPY-513 - Update get_inversion_output to retrieve results from .out file instead of comments that were removed in simpeg-0.8.0**


domfournier:
Repeat of @benk-mira commits directly to release/0.8.0 instead of waiting for GEOPY-500

PR #336, opened 2022-05-12
GEOPY-508: Refactor apps for read-only

domfournier:
**GEOPY-508 - Refactor apps for read-only**

related to https://github.com/MiraGeoscience/geoh5py/pull/202


tests/run_tests/apps_run_test.py
```diff
@@ -21,54 +24,152 @@
 # import pytest
 # pytest.skip("eliminating conflicting test.", allow_module_level=True)
 
-project = "./FlinFlon.geoh5"
-
-geoh5 = Workspace(project)
+PROJECT = "./FlinFlon.geoh5"
```
benk-mira, line 27: why the all caps?

tests/run_tests/apps_run_test.py
```diff
@@ -21,54 +24,152 @@
...
+            parent=workspace
+        )
+
+    app = ContourValues(h5file=temp_workspace, plot_result=False)
     app.trigger.click()
 
+    files = os.listdir(path.join(tmp_path, "Temp"))
+    with Workspace(path.join(tmp_path, "Temp", files[0])) as workspace:
+        output = workspace.get_entity("Airborne_TMI")[0]
+        assert output.n_vertices == 2740, "Change in output. Need to verify."
```
benk-mira, line 72: Diggin the extra asserts - much stronger than the plain run()!

benk-mira (commented):
Looks great!  A few comments, but should be a quick back and forth to move on to some testing.

tests/run_tests/apps_run_test.py
domfournier: From PEP8, constants should technically be in caps no?

tests/run_tests/apps_run_test.py
domfournier: indeed. The turns out that calling .run() didn't garantee that it actually ran because the ipywidgets appears to try/except pass if there is an error.

geoapps/base/application.py
```diff
@@ -305,6 +278,7 @@ def get_output_workspace(self, workpath: str = "./", name: str = "Temp.geoh5"):
             name += ".geoh5"
 
         workspace = Workspace(path.join(workpath, name))
+        workspace.close()
```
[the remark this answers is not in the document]
domfournier: Closing the file wipes the class of entities but keeps track of the h5file. As soon as you try to do an operation again on it, it will re-build itself in mode "r+".

geoapps/calculator/application.py
```diff
@@ -93,45 +94,49 @@ def trigger_click(self, _):
...
-        else:
-            xyz = obj.centroids
+        temp_geoh5 = f"{obj.name}_{out_var}_{time():.3f}.geoh5"
+        with self.get_output_workspace(
+            self.export_directory.selected_path, temp_geoh5
+        ) as workspace:
 
-        if out_var not in obj.get_data_list():
-            obj.add_data({out_var: {"values": numpy.zeros(xyz.shape[0])}})
+            obj = obj.copy(parent=workspace)
```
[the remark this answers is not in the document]
domfournier: Yes, the workspace gets re-opened as soon as you try do to an operation read/write. The advantage of the context is the __exit__ method that closes the file at the end.

tests/run_tests/apps_run_test.py
domfournier: Changed along with GEOPY-533

benk-mira:
Apps warn that monitoring directory is inactive even though the live link works:
image

benk-mira:
Potentially related to above... If monitoring directory is enabled and default path left in place, the live link works anyway.  In this case logging provides path for saved file, but doesn't exist:
image
image

benk-mira:
Cannot run inversion from drag and drop ui.json

PR #339, opened 2022-05-18
Geopy 502

domfournier:
**GEOPY-502 - Cannot run peak finder twice as it gives association validation error.**


geoapps/base/application.py
```diff
@@ -301,10 +302,15 @@ def get_output_workspace(self, workpath: str = "./", name: str = "Temp.geoh5"):
...
+        timestamp = f"_{time.time():.0f}"
         if not name.endswith(".geoh5"):
-            name += ".geoh5"
+            name += timestamp + ".geoh5"
+        else:
+            name.replace(".geoh5", timestamp + ".geoh5")
 
-        workspace = Workspace(path.join(workpath, name))
+        h5file = path.join(workpath, name)
+        h5file = existing_file_incrementer(h5file)
```
domfournier, line 312: I don't think we will ever need this since we are adding a timestamp...

geoapps/utils/string.py
```diff
@@ -0,0 +1,32 @@
...
+import re
+
+
+def existing_file_incrementer(filename):
+    """Add an increment of the form (n) to the filename."""
+
+    if os.path.exists(filename):
+
+        root, file = filename.rsplit(os.path.sep, 1)
+        name, extension = file.split(".", 1)
```
domfournier, line 18: This will return the first occurance, which wouldn't work with the time stamp as it is used now.
i.e. `"123.456.789".split(".", 1)` returns `["123", "456.789"]`

domfournier (commented):
Let's hold off on this until GEOPY-508 is merged. It should fix the issue.

geoapps/base/application.py
benk-mira: What issue will the timestamp come in with though... I tested 508 yesterday and was still getting double run errors

PR #340, opened 2022-05-25
GEOPY-478: Scatter plots driver/uijson/constants

jamiefb:


benk-mira:
Just tested and it looks really good.  I just have a few comments:
1. These should be floats, not integers:
image
2. We should add some sort of logging to the run:
image
3. Saved html is blank:
image
4. Think we should bring the min max defaults over from the app:
image

domfournier (approved):
Very nice!
image

PR #344, opened 2022-06-02
GEOPY-541: Avoid double run of forward

domfournier:
GEOPY-541: Avoid double run of the forward

### DIFF
--- a/geoapps/inversion/driver.py
+++ b/geoapps/inversion/driver.py
@@ -216,6 +216,7 @@ class InversionDriver:
             self.inversion_data.simulate(
                 self.starting_model, self.inverse_problem, self.sorting
             )
+            self.logger.end()
             return
 
         # Run the inversion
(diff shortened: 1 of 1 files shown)


geoapps/inversion/driver.py
```diff
@@ -217,8 +212,10 @@ def run(self):
         """Run inversion from params"""
 
         if self.params.forward_only:
-            self.logger.end()
-            print("Forward simulation completed.")
+            print("Running the forward simulation ...")
+            self.inversion_data.simulate(
+                self.starting_model, self.inverse_problem, self.sorting
+            )
```
sebhmg, line 218: should have `self.logger.end()` here?

sebhmg (changes requested):
missing a logger.end before returning from run()

sebhmg (approved):

PR #347, opened 2022-06-06
GEOPY-539: Create dash interface to Scatter plot

jamiefb:
**GEOPY-539 - Create dash interface to Scatter plot**


benk-mira (dismissed):
Nothing to add here since I'm not well versed in plotly/dash.  Off to testing - let's see how it looks!

PR #348, opened 2022-06-07
GEOPY-530: Change format value for iso-surface and contours applications

jamiefb:
**GEOPY-530 - Change format value for iso-surface and contours applications**


benk-mira (approved):
Approved and moving on to test

PR #350, opened 2022-06-14
GEOPY-539: Create dash interface to Scatter plot

jamiefb:
**GEOPY-539 - Create dash interface to Scatter plot**


domfournier (approved):
Exciting!!

PR #351, opened 2022-06-14
GEOPY-397: Resimulate FlinFlon potential fields data grid

jamiefb:
**GEOPY-397 - Resimulate FlinFlon potential fields data grid (tensor components are wrong orientation)**


benk-mira (changes requested):
I think we'll need to look at this again, the data seem to have changed significantly - I was expecting just a slight shift in the orientation of the anomalies.  Also, the data saved into the workspace is a copy of a copy of a copy, etc... :
image

I've seen this before, but I'm not sure why it occurs.  We'll need to clean that up before we merge.

domfournier:
Handled by #447

PR #352, opened 2022-06-14
GEOPY-519: Chargeability channel not writing from app_initializer

jamiefb:
**GEOPY-519 - chargeability channel not writing from app_initializer in write_default_uijson --use_initializers call.**


geoapps/utils/write_default_uijson.py
```diff
@@ -54,7 +54,9 @@ def write_default_uijson(path, use_initializers=False):
     app_initializer["geoh5"] = path_to_flinflon("FlinFlon_dcip.geoh5")
     dc_init = app_initializer if use_initializers else {}
 
-    from geoapps.inversion.electricals.direct_current.constants import app_initializer
+    from geoapps.inversion.electricals.induced_polarization.constants import (
+        app_initializer,
+    )
```
benk-mira, line 59: Hah, nice catch!

benk-mira (approved):
That should do it!

benk-mira (changes requested):
Looks like there's something wrong with the ip app_initializer that we weren't seeing before since we were pointing to the wrong one.  Here's an image of what I mean with the dc ui.json open to the left for reference
image

PR #353, opened 2022-06-14
GEOPY-458: Allow both "Vertex" and "Cell" values for "topography"

jamiefb:
**GEOPY-458 - Allow both "Vertex" and "Cell" values for "topography"**


benk-mira (approved):
straight forward!

PR #357, opened 2022-06-16
GEOPY-544

benk-mira:
**GEOPY-544 - creating octree_mesh corrupts the input geoh5**


domfournier (dismissed):
Ok so the issue was with update the root on close?

jamiefb (approved):

PR #359, opened 2022-06-21
GEOPY-549: Dash application execution from index page

jamiefb:
**GEOPY-549 - Dash application execution from index page**


environment.yml
```diff
@@ -32,6 +32,7 @@ dependencies:
   - ipyfilechooser>=0.6.0,<0.7.0
   - ipywidgets>=7.6.5,<8.0.0
   - plotly>=5.8.0,<6.0.0
+  - jupyter-dash
```
domfournier, line 35: should we version it too?

PR #364, opened 2022-06-28
GEOPY-559: Cannot write a second ui.json in potential field app

benk-mira:
**GEOPY-559 - Cannot write a second ui.json in potential field app**
Create new params, copy entities over and write file instead of trying to use existing params class.  Applied to both the potential fields and electrical methods applications


geoapps/inversion/potential_fields/application.py
```diff
@@ -1252,12 +1272,27 @@ def write_trigger(self, _):
...
+            ifile.data.update(param_dict)
+
+            new_params = self.params.__class__(input_file=ifile)
+            new_params.write_input_file(
                 name=self._ga_group_name.value + ".ui.json",
                 path=self.export_directory.selected_path,
             )
+            ifile.name = self._ga_group_name.value + ".ui.json"
+            ifile.path = self.export_directory.selected_path
+            self.params.input_file = ifile
```
domfournier, line 1294: Just a note that I will make changes to avoid this. We don't want to replace the self.params while the application is running as it would be pointing to a new workspace with less entities. This would have consequences on the widgets.

domfournier (approved):

PR #365, opened 2022-06-28
GEOPY-557: Update iso-surface trigger_click

domfournier:
**GEOPY-557 - Update iso-surface trigger_click**


sebhmg:
tests expected to pass after the reference to geoh5py dependency is updated. See https://github.com/MiraGeoscience/geoapps/pull/367

benk-mira (dismissed):
Looks good

PR #366, opened 2022-06-28
GEOPY 531: Driver/ui.json/constant for clustering application

jamiefb:
**GEOPY-531 - Driver/ui.json/constant for clustering application**


domfournier (commented):
looking great, but I got a crash when setting the color axis to `kmeans`?
image

domfournier:
@jamiefb I pushed a couple minor import fixes for the latest geoh5py, and to suppress the kmeans warnings I was getting here. Just need to pull again before fixing

domfournier (dismissed):

PR #368, opened 2022-06-29
GEOPY-560: Octree mesh and peak finder apps overwrite the ui.json every time

benk-mira:
**GEOPY-560 - Octree mesh and peak finder apps overwrite the ui.json every time and file isn't named with matching timestamp of the .geoh5 file.**
Fix ui.json writing behaviour from octree and peak finder apps


jamiefb (approved):
Looks good :D

PR #382, opened 2022-07-12
GEOPY 579: Create block model application

jamiefb:
**GEOPY-579 - Create block model application**


geoapps/grid_creation/application.py
```diff
@@ -0,0 +1,383 @@
...
+            update_dict[trigger + "_name"] = locals()[trigger]
+        elif trigger != "":
+            update_dict[trigger] = locals()[trigger]
+
+        self.update_param_dict(update_dict)
+        outputs = self.get_outputs(param_list, update_dict)
+
+        return outputs
+
+    def create_block_model(self, n_clicks):
```
benk-mira, line 335: Doesn't look like n_clicks is used in the function body.  Also the rest of the applications would call this method trigger_click.  I'm tempted to try to keep it consistent with the rest of the apps, but this is a more descriptive name so maybe better to update the other apps?  @domfournier  what do you think?

geoapps/grid_creation/driver.py
```diff
@@ -0,0 +1,85 @@
...
+
+if __name__ == "__main__":
+    print("Loading geoh5 file . . .")
+    file = sys.argv[1]
+    ifile = InputFile.read_ui_json(file)
+    params = GridCreationParams(ifile)
+    driver = GridCreationDriver(params)
+    print("Loaded. Creating block model . . .")
+    driver.run()
+    print("Saved to " + ifile.path)
```
benk-mira, line 85: use path_name here to get the file name in the string

geoapps/grid_creation/driver.py
```diff
@@ -0,0 +1,85 @@
...
+        pads = [
+            self.params.horizontal_padding,
+            self.params.horizontal_padding,
+            self.params.horizontal_padding,
+            self.params.horizontal_padding,
+            self.params.bottom_padding,
+            0.0,
+        ]
+
+        object_out = DataInterpolation.get_block_model(
```
benk-mira, line 50: Seems like the get_block_model method should be stored in the block model application and called by the interpolation app, not the other way around.

assets/uijson/grid_creation.ui.json
```diff
@@ -0,0 +1,86 @@
...
+            "{202C5DB1-A56D-4004-9CAD-BAAFD8899406}",
+            "{6A057FDC-B355-11E3-95BE-FD84A7FFCB88}",
+            "{F26FEBA3-ADED-494B-B9E9-B2BBCBE298E1}",
+            "{4EA87376-3ECE-438B-BF12-3479733DED46}",
+            "{48f5054a-1c5c-4ca4-9048-80f36dc60a06}"
+        ],
+        "label": "Input object",
+        "value": ""
+    },
+    "xy_reference": {
```
benk-mira, line 27: Make this optional so that the user can provide one object that gives both vertical and horizontal extents.

benk-mira (changes requested):
Thanks for this.  I've got a few comments regarding how the code integrates with the existing geoapps.  I'm hoping in a couple cases that we can borrow existing objects/methods to reduce duplication, but let me know if they simply can't be bent into the shape you need them!

geoapps/grid_creation/application.py
domfournier: Can you just pass `_` if you need a placeholder for the callback.

assets/uijson/grid_creation.ui.json
```diff
@@ -0,0 +1,86 @@
+{
+    "title": "Block Model Creation",
+    "geoh5": "",
+    "run_command": "geoapps.grid_creation.driver",
```
domfournier, line 4: Can we rename this `block_model_creation`? Will be more obvious the type of objects we are targeting.

geoapps/grid_creation/application.py
```diff
@@ -0,0 +1,383 @@
...
+        cell_size_x,
+        cell_size_y,
+        cell_size_z,
+        depth_core,
+        horizontal_padding,
+        bottom_padding,
+        expansion_fact,
+        live_link,
+        monitoring_directory,
+    ):
```
domfournier, line 278: Don't forget docstrings

geoapps/grid_creation/application.py
```diff
@@ -0,0 +1,383 @@
...
+        with ws as workspace:
+            # Put entities in output workspace.
+            param_dict["geoh5"] = workspace
+            for key, value in param_dict.items():
+                if isinstance(value, ObjectBase):
+                    param_dict[key] = value.copy(parent=workspace, copy_children=True)
+
+            # Write output uijson.
+            ifile = InputFile(
+                ui_json=self.params.input_file.ui_json,
```
domfournier, line 365: Remove the ifile altogether

geoapps/grid_creation/application.py
```diff
@@ -0,0 +1,383 @@
...
+                validation_options={"disabled": True},
+            )
+            new_params = GridCreationParams(input_file=ifile, **param_dict)
```
domfournier, line 368: Call without the input_file argument. That way params will have a clean input file

geoapps/grid_creation/driver.py
```diff
@@ -0,0 +1,85 @@
...
+            monitored_directory_copy(
+                os.path.abspath(self.params.monitoring_directory), object_out
+            )
+
+
+if __name__ == "__main__":
+    print("Loading geoh5 file . . .")
+    file = sys.argv[1]
+    ifile = InputFile.read_ui_json(file)
+    params = GridCreationParams(ifile)
```
domfournier, line 81: add `params.geoh5.close()`, then open("r+") in inside the run. As to do with input file being read-only going forward.

geoapps/block_model_creation/driver.py
```diff
@@ -0,0 +1,193 @@
...
+from geoh5py.ui_json import InputFile, monitored_directory_copy
+from geoh5py.workspace import Workspace
+from scipy.spatial import cKDTree
+
+from geoapps.block_model_creation.params import BlockModelParams
+from geoapps.shared_utils.utils import get_locations
+
+
+class BlockModelDriver:
+    def __init__(self, params: BlockModelParams):
```
domfournier, line 26: Add a quick docstring description

geoapps/block_model_creation/driver.py
```diff
@@ -0,0 +1,193 @@
...
+            below_core_ind, 2
+        ] = core_bottom_elev  # sets locations below core to core bottom
+        return locs
+
+    @staticmethod
+    def minimum_depth_core(
+        locs: np.ndarray, depth_core: int, core_z_cell_size: int
+    ) -> float:
+        """
+        Get minimum depth core.
```
domfournier, line 50: Space

geoapps/block_model_creation/driver.py
```diff
@@ -0,0 +1,193 @@
...
+        zrange = locs[:, 2].max() - locs[:, 2].min()  # locs z range
+        if depth_core >= zrange:
+            return depth_core - zrange + core_z_cell_size
+        else:
+            return depth_core
+
+    @staticmethod
+    def find_top_padding(obj: BlockModel, core_z_cell_size: int) -> float:
+        """
+        Loop through cell spacing and sum until core_z_cell_size is reached.
```
domfournier, line 65: Space

geoapps/block_model_creation/application.py
```diff
@@ -0,0 +1,486 @@
...
+                                "display": "inline-block",
+                                "margin_bottom": "20px",
+                            },
+                        ),
+                    ]
+                ),
+                html.Div(
+                    [
+                        dcc.Markdown(
+                            children="Minimum x cell size",
```
domfournier, line 119: Somehow the name get's displayed at the end?
image

geoapps/block_model_creation/constants.py
```diff
@@ -0,0 +1,139 @@
...
+    "output_path": os.path.abspath("../../assets/Temp"),
+    "objects": "{2e814779-c35f-4da0-ad6a-39a6912361f9}",
+    "xy_reference": "{2e814779-c35f-4da0-ad6a-39a6912361f9}",
+    "cell_size_x": 50.0,
+    "cell_size_y": 50.0,
+    "cell_size_z": 50.0,
+    "depth_core": 500,
+    "expansion_fact": 1.05,
+    "new_grid": "BlockModel",
+    "horizontal_padding": 0.0,
```
domfournier, line 137: Let's add some padding for defaults, say 500

benk-mira (changes requested):
When I reload from a ui.json file, the output path should take on the path of the file, but it defaults to ./ which creates the file within the geoapps repo:
image

benk-mira:
There's an extra temp file being created when we run the driver:
image

PR #383, opened 2022-07-13
GEOPY-556: Make all jupyter apps read-only

domfournier:
**GEOPY-556 - Make all jupyter apps read-only**


benk-mira (dismissed):
Looks good!

PR #387, opened 2022-07-15
GEOPY 531: Driver/ui.json/constant for clustering application

jamiefb:
**GEOPY-531 - Driver/ui.json/constant for clustering application**


geoapps/clustering/plot_data.py
```diff
@@ -0,0 +1,12 @@
+#  Copyright (c) 2022 Mira Geoscience Ltd.
+#
+#  This file is part of geoapps.
+#
+#  geoapps is distributed under the terms and conditions of the MIT License
+#  (see LICENSE file at the root of this source code package).
+
+
+class PlotData:
```
benk-mira, line 9: Cleanup?

benk-mira (dismissed):
Looks like there's a file that needs cleaning out.  Otherwise looks good.  I'm starting to wonder if we can organize the dash apps a little better. I think it might be nice to split out the callbacks and layout stuff into their own files, this should (roughly) give us a model-view-controller design.  Anyway, not required for this PR, but it's probably a good moment to think overall structure now that we have a couple of apps that have been dashed!

benk-mira:
Can create a callback that updates the options in the cluster color picker to be 1-n_clusters?

benk-mira:
What's this for?:
image

benk-mira:
Can we use a different port to allow more than one instance to run at the same time?:
image

benk-mira:
Data list should contain only the channels selected from the data subset list:
image

domfournier (approved):

PR #389, opened 2022-07-21
GEOPY-578

benk-mira:
**GEOPY-578 - Isosurfaces app raises AssociationValidationError when refreshing values after loading a .ui.json file.**
Fixes a workspace is closed error rather than the bug from the original issue


geoapps/peak_finder/driver.py
```diff
@@ -331,6 +331,5 @@ def rotation_2D(angle):
 if __name__ == "__main__":
     file = sys.argv[1]
     params = PeakFinderParams(InputFile.read_ui_json(file))
-    params.geoh5.close()
```
domfournier, line 334: I think we have to close it here, otherwise the run will try to re-open in r+

geoapps/iso_surfaces/driver.py
```diff
@@ -34,6 +34,7 @@ def run(self):
         Create iso surfaces from input values
         """
 
+        workspace = self.params.geoh5.open(mode="r+")
```
domfournier, line 37: We have a choice here to assume the workspace was opened in r+ on params before calling driver.run?

geoapps/peak_finder/driver.py
benk-mira: That was the case until we made the read_ui_json method close the file.

domfournier (approved):

domfournier (approved):

PR #393, opened 2022-08-04
GEOPY 588: Use base/dash_application with scatter

jamiefb:
**GEOPY-588 - Use base/dash_application with scatter**


geoapps/base/dash_application.py
```diff
@@ -71,93 +190,134 @@ def get_outputs(param_list: list, update_dict: dict) -> tuple:
...
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+                in_use = s.connect_ex(("localhost", p)) == 0
+            if in_use is False:
+                port = p
+                break
+        if port is None:
+            print("No open port found.")
+        return port
 
     def run(self):
```
domfournier, line 309: Could try to open the workspace in read-only here as a context...

domfournier (commented):
Let's try to move the `self.workspace.open(mode='r')` in the run, and remove from the callbacks... might help us for speed. See comments below.

domfournier:
Can we close this PR and let GEOPY-592 carry it all forward?

PR #394, opened 2022-08-09
GEOPY 592: Use base/dash_application for cluster app

jamiefb:
**GEOPY-592 - Use base/dash_application  for cluster app**


domfournier:
I did the merge with 0.8.1, so we should try to fix the new pylint warnings that crop up here. @jamiefb  Mind having a shot at it?

domfournier (approved):
Thanks for the changes

PR #398, opened 2022-08-22
GEOPY-609

domfournier:
**GEOPY-609 - Do not delete topography object**


benk-mira (dismissed):
Looks good to me.  Not sure why we went out of our way to remove in the first place..

PR #401, opened 2022-09-06
Geopy 563

domfournier:
**GEOPY-563 - DC app loaded from ui.json doesn't carry over the mesh parameters**


benk-mira (approved):
Lookin good!

PR #413, opened 2022-09-29
GEOPY-621: Run batch 2D inversion using Sweeper class

benk-mira:
**GEOPY-621 - Run batch 2D inversion using Sweeper class**


geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,120 @@
...
+        for child in survey.children:  # fill a chunk of values from one line
+            if "Iteration" in child.name:
+                if ind is None:
+                    ind = ~np.isnan(child.values)
+                data[child.name]["values"][ind] = child.values[ind]
+
+        return data
+
+
+if __name__ == "__main__":
```
domfournier, line 111: Would be great if we could still call this inversion with the `inversion.driver` instead, like the others. Feels like those methods should be part of the Sweeper class, which could append its methods to the `InversionDriver`.

domfournier: And looks like GEOPY-620 needs come in first, because `sweep` doesn't exist publicly yet.

benk-mira: yes, let's defer to next sprint

benk-mira: Removed the pseudo3d driver

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/constants.py
```diff
@@ -0,0 +1,338 @@
...
+    "starting_model": {
+        "association": "Cell",
+        "dataType": "Float",
+        "group": "Starting Model",
+        "main": True,
+        "isValue": False,
+        "parent": "starting_model_object",
+        "label": "Conductivity (Siemens/m)",
+        "property": None,
+        "value": 0.0,
```
domfournier, line 237: Hey would you mind changing this here and for the 3D case, zero is bad for conductivity

domfournier (approved):
Minor change for start value

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/constants.py
benk-mira: the real value (1e-1) from the defaults dictionary overrides this values anyway, but I've replaced everywhere in the dc packages just for show.

geoapps/inversion/constants.py
```diff
@@ -154,8 +154,6 @@
         "group": "Mesh and Models",
         "main": True,
         "enabled": True,
-        "dependency": "mesh",
```
domfournier, line 157: Should we not remove those meshing parameters altogether?

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+    def write_files(self, lookup):
+        """Write ui.geoh5 and ui.json files for sweep trials."""
+
+        forward_only = self.pseudo3d_params.forward_only
+        ifile = DirectCurrent2DParams(forward_only=forward_only).input_file
+
+        with self.workspace.open(mode="r+"):
+
+            window = InversionWindow(self.workspace, self.pseudo3d_params)
+            data = InversionData(self.workspace, self.pseudo3d_params, window.window)
```
domfournier, line 55: For now, until we refactor, we should probablt assigne to `self.data` and `self.window`  to reflect what's happening in the BaseInversionDriver?

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+                        "reference_model": self.pseudo3d_params.reference_model,
+                        "lower_bound": self.pseudo3d_params.lower_bound,
+                        "upper_bound": self.pseudo3d_params.upper_bound,
+                    }
+                )
+
+            for uuid, trial in lookup.items():
+
+                status = trial.pop("status")
+                if status == "pending":
```
domfournier, line 71: Just for clarity, can we just check for the non-pending case and continue instead. Will avoid the super long indentation.

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+                                {name: {"values": model_values}}
+                            )
+                            model_uids[name] = model_object.uid
+
+                        for key in ifile.data:
+                            param = getattr(self.pseudo3d_params, key, None)
+                            if hasattr(param, "uid"):
+                                if not isinstance(param, Data):
+                                    param.copy(
+                                        parent=iter_workspace, copy_children=True
```
domfournier, line 124: Why do we need to copy over all the originals? Feels like topo is enough here.

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+                            ifile.data[key] = param
+
+                        ifile.data.update(
+                            dict(
+                                lookup[uuid],
+                                **{
+                                    "title": ifile.data["title"].replace("batch ", ""),
+                                    "inversion_type": ifile.data[
+                                        "inversion_type"
+                                    ].replace("pseudo 3d", "2d"),
```
domfournier, line 136: ifile is already 2D, don't think we need any of the text replacements.

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+                                **model_uids,
+                            )
+                        )
+
+                    ifile.name = f"{uuid}.ui.json"
+                    ifile.path = self.working_directory
+                    ifile.write_ui_json()
+                    lookup[uuid]["status"] = "written"
+                else:
+                    lookup[uuid]["status"] = status
```
domfournier, line 150: I don't think this does anything since you just pop and re-insert. Consider the simpler logic suggested above.

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -0,0 +1,152 @@
...
+
+            self.workspace.remove_entity(
+                [k for k in self.workspace.groups if isinstance(k, SimPEGGroup)][0]
```
domfournier, line 153: Why?

geoapps/inversion/line_sweep/driver.py
```diff
@@ -0,0 +1,110 @@
...
+    def line_files(path):
+        with open(os.path.join(path, "lookup.json"), encoding="utf8") as file:
+            line_files = {v["line_id"]: k for k, v in json.load(file).items()}
+        return line_files
+
+    def collect_results(self):
+        path = os.path.join(os.path.dirname(self.workspace.h5file))
+        files = LineSweepDriver.line_files(path)
+        lines = np.unique(self.pseudo3d_params.line_object.values)
+        results_group = SimPEGGroup.create(self.workspace, name="Pseudo3DInversion")
```
domfournier, line 81: Related to the why? above, technically the driver should already have its group.

geoapps/inversion/constants.py
benk-mira: yeah... maybe the octree ones, but we need the ones that are used in the pseudo3d ui.json

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: done

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: done

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: We also need the data object, but I can do the two call separately from the loop. Done

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: I ended up with the title and inversion_type of the batch 2d from the self.pseudo_params update loop.  I've bypassed those to params within the loop rather than correcting after

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: done

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
benk-mira: Not needed anymore,  cleaned up the .constants file

geoapps/inversion/line_sweep/driver.py
benk-mira: done

domfournier (approved):
Looking good. Let's merge geoh5py and sweeper 621, then re-point the yml files before merging.

domfournier (approved):
DismalColorfulBeardedcollie-size_restricted

PR #418, opened 2022-10-26
Geopy 567

domfournier:
**GEOPY-567 - Adding same object in more than one dropdown in octree mesh app and writing ui.json/ui.geoh5 results in duplicate objects in geoh5**


benk-mira (dismissed):
all good

PR #440, opened 2022-11-09
Release/0.9.0 post

domfournier:


sebhmg:
post branch is only to deal with versioning in PyPI as 0.9.0 had to be removed from PyPI and the version cannot be used anymore.
But development still goes on release/0.9.0

andrewg-mira (commented):
Version may need to be post.2 because I mistakenly pushed a 0.9.0-post.1 to pypi yesterday.

PR #441, opened 2022-11-09
GEOPY-694

sebhmg:
**GEOPY-694 - Crash on MVI with curve entity**


sebhmg:
same as https://github.com/MiraGeoscience/geoapps/pull/438, but targeting the release branch

PR #451, opened 2022-12-01
GEOPY-710

domfournier:
**GEOPY-710 - Fix distributed_workers validation**


benk-mira:
Using other branch updated with release-0.9.1

PR #463, opened 2022-12-15
GEOPY-727: Detection of disconnected air cells don't trigger on 2D meshes

domfournier:
**GEOPY-727 - Detection of disconnected air cells don't trigger on 2D meshes**


benk-mira (approved):
Looks good

benk-mira (approved):
looks good

PR #464, opened 2022-12-15
GEOPY-732: Workspace closed on ui.json execution of non-inversion apps

domfournier:
**GEOPY-732 - Workspace closed on ui.json execution of non-inversion apps**


benk-mira (approved):
Seems like that should have broken the inversions too, since we're writing iterations to the geoh5...

PR #471, opened 2023-01-04
GEOPY-823: tests can use projects in assets

sebhmg:
**GEOPY-823 - tests can use directly project in assets instead of download from github**

[GEOPY-823]: https://mirageoscience.atlassian.net/browse/GEOPY-823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/params_test.py
```diff
@@ -1255,7 +1256,7 @@ def test_gy_uncertainty():
 
 mag_params = MagneticScalarParams(
     **{
-        "geoh5": "./FlinFlon.geoh5",
+        "geoh5": str(PROJECT),
```
domfournier, line 1259: To be addressed on GEOPY-830, but Params should be able to take on `Path` type input for `geoh5` and `workspace_geoh5` values

domfournier (approved):
Looks great, thanks a lot. I will test now.

PR #472, opened 2023-01-05
GEOPY-824: use Python 3.10 as default runtime environment

sebhmg:
**GEOPY-824 - use Python 3.10 as default run environment**


sebhmg:
test failures are not related to these changes. They will be fixed by GEOPY-823

domfournier (dismissed):
Pending to test fixes on GEOPY-823

PR #484, opened 2023-01-06
GEOPY-828: pause before exit to show error when conda not found

sebhmg:
**GEOPY-828 - new installation location for latest miniconda version**


sebhmg:
now, the command windows stays in case of error instead of closing right away:

image

domfournier (approved):

PR #488, opened 2023-01-12
GEOPY-835

domfournier:
**GEOPY-835 - Iso-surface creation jupyter-app error after crash of marhcing cube**
…o float parameter

### DIFF
--- a/geoapps/iso_surfaces/driver.py
+++ b/geoapps/iso_surfaces/driver.py
@@ -17,6 +17,7 @@ from geoh5py.objects import BlockModel, ObjectBase, Surface
 from geoh5py.ui_json.utils import monitored_directory_copy
 from scipy.interpolate import interp1d
 from skimage.measure import marching_cubes
+from tqdm import tqdm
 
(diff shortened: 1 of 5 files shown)


geoapps/iso_surfaces/driver.py
```diff
@@ -158,9 +160,13 @@ def iso_surface(
...
             values = values[0].reshape(x.shape)
 
+        lower, upper = np.nanmin(values), np.nanmax(values)
         surfaces = []
-        for level in levels:
+        print("Running marching cubes on levels.")
+        for level in tqdm(levels):
             try:
+                if level < lower or level > upper:
+                    continue
```
benk-mira, line 169: Maybe log that the levels are being skipped?

domfournier: Good call. Done

benk-mira: That should do it!

benk-mira (approved):
cool

PR #500, opened 2023-01-24
GEOPY-585

benk-mira:
**GEOPY-585 - Inversion apps point to bogus .geoh5 since naming has been timestamped**


domfournier (approved):
Good idea. I ll test it now.

PR #501, opened 2023-01-31
GEOPY-585

domfournier:
**GEOPY-585 - Inversion apps point to bogus .geoh5 since naming has been timestamped**

### DIFF
--- a/geoapps/utils/plotting.py
+++ b/geoapps/utils/plotting.py
@@ -7,7 +7,9 @@
 
 from __future__ import annotations
 
+import warnings
 from copy import copy
+from os import path
 
 import numpy as np
 from geoh5py.data import Data
@@ -540,8 +542,19 @@ def check_data_type(data):
     return data
 
 
-def plot_convergence_curve(outfile):
-    """"""
+def plot_convergence_curve(outfile: str) -> widgets.interactive | None:
+    """
+    Plot the convergence curve from a SimPEG *.out file.
+
+    :param outfile: Full path to the SimPEG *.out file
+
+    :return fig: widgets.interactive figure.
+    """
+
+    if not path.exists(outfile):
+        warnings.warn(f"File {outfile} does not exist.")
+        return
+
     with open(outfile, encoding="utf-8") as f:
         lines = f.readlines()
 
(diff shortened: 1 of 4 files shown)


geoapps/utils/plotting.py
```diff
@@ -541,38 +542,73 @@ def check_data_type(data):
...
+    """
+
+    if not path.exists(outfile):
+        warnings.warn(f"File {outfile} does not exist.")
+        return
+
+    with open(outfile, encoding="utf-8") as f:
+        lines = f.readlines()
+
+    names = lines[0].split(" ")[:-1]
```
benk-mira, line 561: why not just use np.genfromtxt here.  Seems like it should be flexible enough..

geoapps/utils/plotting.py
```diff
@@ -541,38 +542,73 @@ def check_data_type(data):
...
-            ax1 = plt.subplot()
-            ax2 = ax1.twinx()
-            ax1.plot(iterations, phi_d, linewidth=3, c="k")
-            ax1.set_xlabel("Iterations")
-            ax1.set_ylabel(r"$\phi_d$", size=16)
-            ax2.plot(iterations, phi_m, linewidth=3, c="r")
-            ax2.set_ylabel(r"$\phi_m$", size=16)
+        ax1 = plt.subplot()
+        ax2 = ax1.twinx()
+        ax1.plot(iterations, phi_d, linewidth=3, c="k")
```
benk-mira, line 598: I'm not following you here.  We always plot phi_d and phi_m with the labels set accordingly, but phi_d and phi_m take the data from the dropdown widgets which include iteration, beta as options.  Why would we want to plot iterations/beta vs iterations and label it phi_d/phi_m??  Shouldn't we just drop the widgets and always plot phi_d and phi_m?  Or have a widget that switches between a phid/phim plot and a beta vs iter plot?

benk-mira (changes requested):
Confused by the new plotting code..

geoapps/utils/plotting.py
domfournier: It was so that users could plot phi_d/phi_m (the classic elbow). But granted, the labels should be changing ...
image

geoapps/utils/plotting.py
domfournier: Good call. I had tried `numpy.loadtxt` and that failed because of the time-date column. Will switch over.

benk-mira: genfromtxt should have a usecols to skip the dates

domfournier: Done

benk-mira (approved):
Great!

PR #502, opened 2023-02-01
GEOPY-837

domfournier:
**GEOPY-837 - Fix import warnings for inversion**


tests/run_tests/driver_mt_test.py
```diff
@@ -25,8 +25,8 @@
 
 target_run = {
     "data_norm": 0.01577,
-    "phi_d": 12.04,
-    "phi_m": 0.2547,
+    "phi_d": 11.52,
```
benk-mira, line 28: I guess these are changing because something was updated in the new SimPEG version we're pointing to in the pyproject.toml?  What changed?

benk-mira (commented):
Looks like pointing to a different SimPEG verstion in the pyproject.toml affects the solutions to the mt/mvi run tests.  Did you verify that the full solutions still make sense?

PR #504, opened 2023-02-11
Change the log file naming

aluthfian:
I would love to see my log file not being continuously overwritten during batch inversion.

Thanks, Fian


aluthfian:
I used renaming capabilities from batch file so all good now.

Thanks, Fian

PR #510, opened 2023-02-21
GEOPY-858: restrict conda channels to conda-forge

sebhmg:
**GEOPY-858 - restrict conda channels to conda-forge**


andrewg-mira (approved):

conda-py-3.10-lock.yml
```diff
@@ -11,26 +11,24 @@
...
+    win-64: b3629c7344970ad90c6c4730fe32e887c542dac6edf46b3d18398847def9cab3
   platforms:
   - win-64
   - osx-64
   - linux-64
   sources:
-  - pyproject.toml
-  - environments/env-python-3.10.yml
+  - /mnt/c/Users/sebastienh/Dev/geoapps/pyproject.toml
+  - /mnt/c/Users/sebastienh/Dev/geoapps/environments/env-python-3.10.yml
```
benk-mira, line 31: That's probably not supposed to be there!

benk-mira (changes requested):
I'm not a great reviewer for the .bat stuff, but I didn't see anything obviously wrong.  I did notice a path with your local directory in one of the lock.yml files - could be in more (didn't inspect all of them).

devtools/run_conda_lock.py
```diff
@@ -91,6 +96,28 @@ def per_platform_env(py_ver: str, extras=[], dev=False, suffix="") -> None:
             f.write(env_file_variables_section_)
 
 
+def patch_absolute_path(file: Path) -> None:
+    """
+    Patch the given file to remove reference with absolute file path.
+    """
+
+    abs_path_base = str(_environments_folder.absolute().parent) + os.sep
+    assert abs_path_base.endswith(os.sep)
```
benk-mira, line 105: This is redundant since you just added it to the string.  Also, os.path has some convenience functions for working with path strings.  I use os.path.join a lot which will join strings with the os.sep by default and will only use one separator in case one of the strings already has one.  Might be useful here.

benk-mira (approved):
Nothing blocking, just some suggestions for path manipulations

PR #515, opened 2023-02-24
GEOPY-866: locked env yml contains hash for pip package that make the installation fail

sebhmg:
**GEOPY-866 - locked env yml contains has for pip package that make the installation fail**


benk-mira (approved):
All good here, on to test..

PR #519, opened 2023-03-22
GEOPY-883: more robust dev, install and run scripts

sebhmg:
**GEOPY-883 - conda not found at install time**

### DIFF
--- a/Start_applications.bat
+++ b/Start_applications.bat
@@ -7,6 +7,8 @@ if !errorlevel! neq 0 (
   exit /B !errorlevel!
 )
 
+set ENV_NAME=geoapps
+
 set MY_CONDA=!MY_CONDA_EXE:"=!
-call "!MY_CONDA!" activate geoapps && jupyter notebook geoapps\index.ipynb
-cmd /k
+call "!MY_CONDA!" run -n %ENV_NAME% jupyter notebook geoapps\index.ipynb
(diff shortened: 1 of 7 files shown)


andrewg-mira (approved):

Start_applications.bat
```diff
@@ -7,6 +7,8 @@ if !errorlevel! neq 0 (
   exit /B !errorlevel!
 )
 
+set ENV_NAME=geoapps
+
 set MY_CONDA=!MY_CONDA_EXE:"=!
-call "!MY_CONDA!" activate geoapps && jupyter notebook geoapps\index.ipynb
-cmd /k
+call "!MY_CONDA!" run -n %ENV_NAME% jupyter notebook geoapps\index.ipynb
+cmd /k "!MY_CONDA!" activate %ENV_NAME%
```
domfournier, line 14: Same comment as on GEOPY-891, kind of weird that you activate after.

domfournier: Does that work?

sebhmg: yes it does work. Actually, if jupyter starts, it never reaches that point.
If juyter fails, then it places the user in a terminal where the environment is activated. Possibly useful for diagnostic

domfournier: I see. Great.

domfournier (approved):

PR #524, opened 2023-03-27
GEOPY-891: allow installation with different options: only geoh5+omf, apps no-ui, all

sebhmg:
**GEOPY-891 - allow installation with different options: only geoh5+omf, apps no-ui, all**


domfournier (approved):

domfournier (changes requested):
Ooops, typo here:
https://github.com/MiraGeoscience/geoapps/blob/0014829b4852cde955da381e7b80028089958377/Install_or_Update.bat#L17

Should be `win-64`

domfournier:
And looks like we might be missing pip from the reqs?
image

domfournier:
And get an error running the lock file script
image

domfournier:
Tried to run without the pip install as is, also returns this:

image

PR #526, opened 2023-03-29
GEOPY-720: Interface between uijson and inversion group metadata

domfournier:
**GEOPY-720 - Interface between uijson and inversion group metadata**


benk-mira (dismissed):
Looks good

PR #530, opened 2023-04-04
GEOPY-904

benk-mira:
**GEOPY-904 - Set uncertainties to infinity for nan data**


benk-mira:
Turns out we already handled this right before handing off to SimPEG survey object.  I tried out setting the infinity uncertainty earlier in the DataInversion class, but this caused the dpred to be blank for the nan data.  I reverted the changes so that the observed can have empty data, but the dpred simulates at the missing stations anyway.

PR #535, opened 2023-04-11
GEOPY-920

domfournier:


benk-mira (dismissed):
Thanks for the fix!

geoapps/iso_surfaces/constants.py
```diff
@@ -86,7 +86,7 @@
         "fixed_contours": {
             "main": True,
             "label": "Fixed Contours",
-            "value": 0.0,
+            "value": "",
```
sebhmg, line 89: **question:** is it expecting at runtime a string with coma separated values?

sebhmg (dismissed):

geoapps/iso_surfaces/constants.py
domfournier: yes, but input_file converts the string to None if empty.

PR #540, opened 2023-04-26
GEOPY-934

domfournier:
**GEOPY-934 - Mesh missed placed for non-straight dc lines**

### DIFF
--- a/geoapps/inversion/components/data.py
+++ b/geoapps/inversion/components/data.py
@@ -20,11 +20,12 @@ from copy import deepcopy
 
 import numpy as np
 from discretize import TreeMesh
+from scipy.spatial import cKDTree
 from SimPEG import maps
 from SimPEG.electromagnetics.static.utils.static_utils import geometric_factor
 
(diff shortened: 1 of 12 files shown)


benk-mira (approved):
I see.. I agree that we might need to have another look at this problem with non-straight lines in mind.

geoapps/inversion/components/data.py
```diff
@@ -161,6 +161,28 @@ def _initialize(self) -> None:
         self.survey, _, _ = self.create_survey()
         self.save_data(self.entity)
 
+    def pseudo_locations(self, locations: np.ndarray) -> np.ndarray:
```
benk-mira, line 164: Maybe we should rename to avoid confusion with the dc definition where z is a function of pole/dipole spacings?  How about drape_locations?

benk-mira (dismissed):
Looks good, just a comment about renaming pseudo_locations

geoapps/inversion/components/data.py
domfournier: sounds good

PR #543, opened 2023-05-01
GEOPY-905: Time units not considered by receivers

domfournier:
**GEOPY-905 - Time units not considered by receivers**


benk-mira (approved):
Looks good

PR #546, opened 2023-05-01
GEOPY-927: Allow to run 2D inversion from mesh parameters

domfournier:
**GEOPY-927 - Allow to run 2D inversion from mesh parameters**


benk-mira (approved):
Looks good

PR #550, opened 2023-05-03
GEOPY-934: Mesh missed placed for non-straight dc lines

domfournier:
**GEOPY-934 - Mesh missed placed for non-straight dc lines**


benk-mira (dismissed):
sure!

PR #557, opened 2023-05-10
GEOPY-946

domfournier:


geoapps/inversion/constants.py
```diff
@@ -504,6 +504,15 @@
         "value": True,
         "verbose": 3,
     },
+    "out_group": {
+        "label": "SimPEG group",
+        "value": None,
+        "groupType": "{55ed3daf-c192-4d4b-a439-60fa987fe2b8}",
+        "group": "Drag-and-drop options",
```
benk-mira, line 511: does this need a group name?  There won't be anything in the group with it, will there?

geoapps/inversion/potential_fields/application.py
```diff
@@ -875,7 +875,13 @@ def inversion_type_observer(self, _):
             self._param_class = GravityParams
 
         self.params = self._param_class(validate=False, verbose=False)
-        self.ga_group_name.value = self.params.out_group
+
+        if getattr(self.params, "_out_group", None) is not None:
+            self.ga_group_name.value = self.params.out_group.name
+        else:
+            self.ga_group_name.value = (
```
benk-mira, line 882: Don't think this is needed since you put it in the base params property.  Here and in the other applications.

benk-mira (commented):
Just a couple of potential cleanups

geoapps/inversion/constants.py
domfournier: That's Kris' call. He likes the group better as the optional goes on it.

geoapps/inversion/potential_fields/application.py
domfournier: I just don't want to change the rest of the behaviour now. Otherwise I need to remove those widgets, and that's bigger.

benk-mira (approved):

PR #561, opened 2023-05-12
GEOPY-950

benk-mira:
**GEOPY-950 - Keep last .out file in pseudo3d runs so that convergence curve doesn't fail**
…e all the models in their simpeggroup under the main simpeggroup of the pseuso3d run


domfournier (dismissed):
Looks pretty easy fix, great. Let's f*%# go!

andrewg-mira (approved):

PR #565, opened 2023-05-17
GEOPY- 947

sebhmg:
requires changes in geoh5py https://github.com/MiraGeoscience/geoh5py/pull/347 and param-sweeps https://github.com/MiraGeoscience/param-sweeps/pull/20


geoapps/scatter_plot/application.py
```diff
@@ -476,7 +477,7 @@ def update_plot(
 
         # Update self.params
         param_dict = self.get_params_dict(update_dict)
-        self.params.update(param_dict, validate=False)
+        self.params.update(param_dict)
```
benk-mira, line 480: how did this sneak in here?  Is this merged over from one of Dom's issues?

benk-mira (commented):
Wow didn't know this issue was going to be so expansive!  Thanks for this.  I just had one comment about removing a validate=False.  We generally reserve the validations until read/write, which I think is why the validate=False was in there.

geoapps/scatter_plot/application.py
sebhmg: no idea. This is in the develop branch, which is failing the static analysis too...

sebhmg: ho, we wish this to be restored? I though that wsa a mistake, as pylint colains about it
`geoapps\scatter_plot\application.py:479:8: E1123: Unexpected keyword argument 'validate' in method call (unexpected-keyword-arg)`
See here on develop branch: https://github.com/MiraGeoscience/geoapps/actions/runs/4996027208/jobs/8948668149

benk-mira: Oh, I see - the argument has been removed from the update method

benk-mira (approved):

geoapps/scatter_plot/application.py
domfournier: Indeed, the scatter and clustering apps were broken. We, unfortunately, don't have good tests for the Dash app, so it went unnoticed. I will ask Jamie to look into unit testing those, but I am afraid it's pretty complicated with callbacks...

domfournier (dismissed):
Big one, thanks for this.

PR #566, opened 2023-05-29
GEOPY-958

jamiefb:
**GEOPY-958 - Fix Scatter plot app and add unittest**


geoapps/scatter_plot/application.py
```diff
@@ -287,11 +288,9 @@ def get_channel_bounds(
 
         if channel == "kmeans" and kmeans is not None:
             data = kmeans
-        elif (
-            channel is not None
-            and self.workspace.get_entity(uuid.UUID(channel))[0] is not None
-        ):
-            data = self.workspace.get_entity(uuid.UUID(channel))[0].values
+        elif channel is not None:
+            if self.workspace.get_entity(uuid.UUID(channel))[0] is not None:
```
domfournier, line 292:
```suggestion
        elif channel is not None and self.workspace.get_entity(uuid.UUID(channel))[0] is not None:
```

tests/run_tests/apps_run_test.py
```diff
@@ -191,51 +193,194 @@ def test_clustering(tmp_path: Path):
...
+    ]
+    ga_group_name = "Clusters"
+    monitoring_directory = str(tmp_path)
+    trigger = "export"
+    clusters = {}
+
+    dataframe_dict, mapping, indices = app.update_dataframe(downsampling, data_subset)
+
+    # Set a callback since callback_context is used by run_clustering
+    context_value.set(
```
domfournier, line 249: Oh interesting, so that's how we can trigger value changes from the outside...

domfournier (commented):
Great work, glad to see how we can trigger the callbacks. See small comments

domfournier (approved):
👍

PR #567, opened 2023-06-05
GEOPY-961

jamiefb:
**GEOPY-961 - Use Qt web window for scatter plot.**


geoapps/base/dash_application.py
```diff
@@ -300,3 +313,253 @@ def workspace(self, workspace):
...
+class ObjectSelection:
+    """
+    Dash app to select workspace and object. Creates temporary workspace with the object, and
+    opens a Qt window to run an app.
+    """
+
+    _app_name = None
+    _workspace = None
+    _param_class = None
+    _app_class = None
```
domfournier, line 327: Let's move these inside the init. Those look more like instance attributes.

geoapps/base/dash_application.py
```diff
@@ -300,3 +313,253 @@ def workspace(self, workspace):
...
+
+    def __init__(self, app_name, app_initializer, app_class, param_class, **kwargs):
```
domfournier, line 329: Add typing

geoapps/base/dash_application.py
```diff
@@ -300,3 +313,253 @@ def workspace(self, workspace):
...
+        self.app_name = app_name
+        self.app_class = app_class
+        self.param_class = param_class
```
domfournier, line 332: Add setters and check for type

geoapps/base/dash_application.py
```diff
@@ -300,3 +313,253 @@ def workspace(self, workspace):
...
+
+        :return launch_app_markdown: Empty string since callbacks must have output.
+        """
+
+        triggers = [c["prop_id"].split(".")[0] for c in callback_context.triggered]
+        if "launch_app" in triggers and objects is not None:
+            # Make new workspace with only the selected object
+            obj = self.workspace.get_entity(uuid.UUID(objects))[0]
+
+            temp_geoh5 = "Scatterplot_" + f"{time():.0f}.geoh5"
```
domfournier, line 509: Let's generalize that and imply append a time stamp to the name of the incoming `workspace.h5file`

geoapps/clustering/layout.py
```diff
@@ -10,7 +10,39 @@
...
+                        "width": "65%",
+                        "display": "inline-block",
+                        "margin_bottom": "40px",
+                        "vertical-align": "bottom",
+                    },
+                ),
+            ]
+        ),
+    ]
+)
```
domfournier, line 45: Can we just re-use the base layout instead?

domfournier (commented):
Thanks! 
Getting there. See comments.

geoapps/base/dash_application.py
```diff
@@ -300,3 +313,313 @@ def workspace(self, workspace):
...
+            self._app_class = val
+        else:
+            raise TypeError
+
+    @property
+    def param_class(self) -> type[BaseParams]:
+        """
+        The param class associated with the launched app.
+        """
+        return self._param_class
```
domfournier, line 604: The private attribute should be defined in the init though. I am surprised that mypy didn't complain about this.

geoapps/base/dash_application.py
```diff
@@ -300,3 +313,313 @@ def workspace(self, workspace):
...
+
+    @param_class.setter
+    def param_class(self, val):
+        if issubclass(val, BaseParams):
+            self._param_class = val
+        else:
+            raise TypeError
```
domfournier, line 611:
```suggestion
        if not issubclass(val, BaseParams):
            raise TypeError("Value for attribute `param_class` should be a subclass of :obj:`geoapps.driver_base.BaseParams`")
            
        self._param_class = val
```

domfournier: Can you do similar things for the other setters as well?

domfournier (commented):
Sorry for being picky - couple more things.

domfournier (commented):
Sorry for being picky - couple more things.

domfournier (approved):

PR #570, opened 2023-06-07
GEOPY-546: add 3rd party notice

jamiefb:
**GEOPY-546 - add 3rd party notice**


.pre-commit-config.yaml
```diff
@@ -86,6 +86,7 @@ repos:
     rev: v6.1.2
     hooks:
     -   id: rstcheck
+        exclude: ^THIRD_PARTY_SOFTWARE.rst$
```
sebhmg, line 89: **suggestion:** exclude as well from codespell hook

lower in that file:
```
    -   id: codespell
        exclude: (-lock\.ya?ml|\benvironments/.*\.ya?ml|\.ipynb|^THIRD_PARTY_SOFTWARE\.rst)$
```

sebhmg (changes requested):
should only list the top-level dependencies

sebhmg (approved):

PR #571, opened 2023-06-07
GEOPY-970: Issue clipping out .ui with Path and stem

domfournier:
**GEOPY-970 - Issue clipping out .ui with Path and stem**


geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,21 @@ def run(self):  # pylint: disable=W0221
 
     def setup_params(self):
         h5_file_path = Path(self.workspace.h5file).resolve()
-        ui_json_path = h5_file_path.with_suffix(".ui.json")
-        if not ui_json_path.is_file():
+        ui_json_path = h5_file_path.stem.replace(".ui", "") + ".ui.json"
+        if not (h5_file_path / ui_json_path).is_file():
```
sebhmg, line 43: **suggestion (naming):" either rename to `ui_json_name` or `ui_json_path` with the full path this way:
```suggestion
        ui_json_path = h5_file_path.parent / h5_file_path.stem.replace(".ui", ".ui.json")
        if not (ui_json_path).is_file():
```

geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,21 @@ def run(self):  # pylint: disable=W0221
 
     def setup_params(self):
         h5_file_path = Path(self.workspace.h5file).resolve()
-        ui_json_path = h5_file_path.with_suffix(".ui.json")
-        if not ui_json_path.is_file():
+        ui_json_path = h5_file_path.stem.replace(".ui", "") + ".ui.json"
```
sebhmg, line 42: I do not understand when this was getting a wrong output

`with_suffix` seems to replace as expected. See example below:
image

sebhmg (changes requested):
can you give an example where this was no working? I do understand why it was wrong

geoapps/inversion/line_sweep/driver.py
sebhmg: ho, ok, it does stem, AND replace, I missed that. thus removing 2 extensions.
Could it be then `f"{h5_file_path.stem}.json"` ?

domfournier: guess so yes

geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,23 @@ def run(self):  # pylint: disable=W0221
...
             with self.workspace.open():
                 self.pseudo3d_params.write_input_file(
-                    name=ui_json_path.name,
-                    path=str(ui_json_path.parent),
+                    name=ui_json_path,
+                    path=h5_file_path.parent,
                 )
         generate(
-            str(ui_json_path),
+            h5_file_path.parent / ui_json_path,
```
sebhmg, line 52: **issue:** `ui_json_path` is already the full path
```suggestion
            ui_json_path,
```

geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,23 @@ def run(self):  # pylint: disable=W0221
...
-        if not ui_json_path.is_file():
+        ui_json_path = h5_file_path.parent / h5_file_path.stem.replace(
+            ".ui", ".ui.json"
+        )
+        if not (ui_json_path).is_file():
             with self.workspace.open():
                 self.pseudo3d_params.write_input_file(
-                    name=ui_json_path.name,
-                    path=str(ui_json_path.parent),
+                    name=ui_json_path,
```
sebhmg, line 48: **issue:** should just be the name
```suggestion
                    name=ui_json_path.name,
```

geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,23 @@ def run(self):  # pylint: disable=W0221
 
     def setup_params(self):
         h5_file_path = Path(self.workspace.h5file).resolve()
-        ui_json_path = h5_file_path.with_suffix(".ui.json")
-        if not ui_json_path.is_file():
+        ui_json_path = h5_file_path.parent / h5_file_path.stem.replace(
+            ".ui", ".ui.json"
+        )
```
sebhmg, line 44: **suggestion:** do not remove ".ui" just anywhere in the name but only at the end
```suggestion
        ui_json_path = h5_file_path.parent / (re.sub(r"\.ui$", "", h5_file_path.stem) + ".ui.json")
```

geoapps/inversion/line_sweep/driver.py
```diff
@@ -39,20 +39,23 @@ def run(self):  # pylint: disable=W0221
...
         generate(
-            str(ui_json_path),
+            h5_file_path.parent / ui_json_path,
             parameters=["line_id"],
             update_values={"conda_environment": "geoapps"},
         )
         ifile = InputFile.read_ui_json(
-            h5_file_path.parent / f"{h5_file_path.stem}_sweep.ui.json"
+            h5_file_path.parent
+            / f"{h5_file_path.stem.replace('.ui', '')}_sweep.ui.json"
```
sebhmg, line 58: **suggestion:** do not remove ".ui" just anywhere in the name but only at the end
```suggestion
            h5_file_path.parent / (re.sub(r"\.ui$", "", h5_file_path.stem) + "_sweep.ui.json")
```

sebhmg (changes requested):
see comments:
- for suggestions on removing .ui only at the end of the name
- confusion between path and name it seems

tests/run_tests/driver_dc_p3d_test.py
```diff
@@ -92,7 +92,7 @@ def test_dc_p3d_run(
 ):
     workpath = tmp_path / "inversion_test.geoh5"
```
sebhmg, line 93: **question:**: based on the other tests, should it be .ui.geoh5 here too?

sebhmg (approved):

PR #572, opened 2023-06-08
GEOPY-546: WIP

sebhmg:
**GEOPY-546 - add 3rd party notice**


THIRD_PARTY_LICENSES2.txt
sebhmg, line 1: To be ignored. Was my attempt to auto-generate them...

PR #575, opened 2023-06-14
GEOPY-607

jamiefb:
**GEOPY-607 - Use Base inversion for Mag/Grav application **


geoapps/inversion/potential_fields/magnetic_scalar/application.py
```diff
@@ -0,0 +1,135 @@
...
+            # Magnetic specific
+            State(component_id="inducing_field_strength", component_property="value"),
+            State(
+                component_id="inducing_field_inclination", component_property="value"
+            ),
+            State(
+                component_id="inducing_field_declination", component_property="value"
+            ),
+            prevent_initial_call=True,
+        )(self.write_trigger)
```
domfournier, line 135: It looks like this callback is duplicated in `gravity` and `magnetic_vector`. Any chance we can move it to the base class?

geoapps/inversion/potential_fields/magnetic_vector/constants.py
```diff
@@ -175,12 +179,24 @@
...
+    "alpha_s": 1.0,
+    "length_scale_x": 1.0,
+    "length_scale_y": 1.0,
+    "length_scale_z": 1.0,
+    "s_norm": 0.0,
+    "x_norm": 2.0,
+    "y_norm": 2.0,
+    "z_norm": 2.0,
+    "fix_aspect_ratio": True,
+    "colorbar": False,
```
domfournier, line 199: Should not need any of those. Are you getting an error?

geoapps/inversion/potential_fields/magnetic_vector/constants.py
```diff
@@ -795,16 +829,20 @@
     "by_uncertainty": {"one_of": "uncertainty channel"},
     "bz_channel": {"one_of": "data channel"},
     "bz_uncertainty": {"one_of": "uncertainty channel"},
+    "fix_aspect_ratio": {"types": [bool]},
+    "colorbar": {"types": [bool]},
```
domfournier, line 833: Also should not have to validate those as they should not be part of the ui.json

domfournier (commented):
See comments below. Also, any chance we can re-scale the colors based on the color stretch in GA? The colormap (if any) is stored under the `data.entity_type.color_map` as an array of [value, r, g, b]

image

domfournier (approved):

PR #577, opened 2023-06-22
GEOPY-757: Add Ground TEM large-loop to geoapps

domfournier:
**GEOPY-757 - Add Ground TEM (large-loop) option with runtest**


geoapps/inversion/components/factories/receiver_factory.py
```diff
@@ -57,7 +57,13 @@ def concrete_object(self):
         elif "tdem" in self.factory_type:
             from SimPEG.electromagnetics.time_domain import receivers
 
-            return receivers.PointMagneticFluxTimeDerivative
+            if self.params.data_units == "dB/dt (T/s)":
+                return receivers.PointMagneticFluxTimeDerivative
+            elif self.params.data_units == "B (T)":
+                return receivers.PointMagneticFluxDensity
```
benk-mira, line 63: What about H:
image

benk-mira (commented):
I think you might need to handle an extra case in the receiver factory, have a look at the comment there..

geoapps/inversion/components/factories/receiver_factory.py
domfournier: The next else returns a PointMagneticField

domfournier: The next else returns a PointMagneticField receiver...

benk-mira (approved):

PR #581, opened 2023-06-30
GEOPY-979

jamiefb:
**GEOPY-979 - Peak Finder crashes trying to access data from closed workspace**


domfournier (approved):
Makes sense. Thanks.
Going to testing.

PR #585, opened 2023-07-10
GEOPY-607

jamiefb:
**GEOPY-607 - Use Base inversion for Mag/Grav application **


benk-mira (approved):
sure

domfournier:
I don't get what's happening here. The diff is only 1 file change? I don't see changes from your other commits?

domfournier (approved):
Sounds good. We can keep a clean state of the json here.

PR #592, opened 2023-08-01
GEOPY-1020

domfournier:
**GEOPY-1020 - Update metadata of inversions to point to new entities. **


geoapps/inversion/driver.py
```diff
@@ -279,22 +278,26 @@ def run(self):
         self.logger.start()
         self.configure_dask()
 
+        simpeg_inversion = self.inversion
```
benk-mira, line 281: why the alias?

benk-mira (approved):

PR #594, opened 2023-08-02
Geopy-882 codecov

andrewg-mira:
**GEOPY-882 - geoh5py not uploading to codecov**


domfournier:
@andrewg-mira I think you want to send this to `develop` instead, but it appears that nothing is different. What's the status of this PR?

PR #595, opened 2023-08-03
GEOPY-989

benk-mira:
**GEOPY-989 - Remove the forward_ui_json and update labels in write_default_uijson**


domfournier (commented):
That would mean that we should be able to remove the forward_ui_json in all constants right?

domfournier (approved):
Nice clean up. Makes ma' brain  happy

PR #597, opened 2023-08-04
GEOPY-1027

domfournier:
**GEOPY-1027 - Fix data/predicted type naming**


geoapps/inversion/components/factories/simulation_factory.py
```diff
@@ -197,7 +197,10 @@ def _conductivity_keywords(self, kwargs, mesh, active_cells=None):
     def _tdem_keywords(self, kwargs, receivers, mesh, active_cells=None):
         kwargs = self._conductivity_keywords(kwargs, mesh, active_cells=active_cells)
         kwargs["t0"] = -receivers.timing_mark * self.params.unit_conversion
-        kwargs["time_steps"] = np.round((np.diff(receivers.waveform[:, 0])), decimals=6)
+        kwargs["time_steps"] = (
+            np.round((np.diff(np.unique(receivers.waveform[:, 0]))), decimals=6)
+            * self.params.unit_conversion
```
benk-mira, line 202: Good catch.  I wonder if we should simulate a different unit in the runtest and check that we get the same target value.

domfournier: Good idea.

domfournier: Done

benk-mira (dismissed):
Nice, thanks for the beefed up test!

PR #607, opened 2023-08-31
GEOPY-875: Add warning message if version differ

domfournier:
**GEOPY-875 - Add geoapps version to ui.json file.**


geoapps/driver_base/driver.py
```diff
@@ -97,6 +99,14 @@ def start(cls, filepath: str | Path, driver_class=None):
             filepath, validations=driver_class._validations  # pylint: disable=W0212
```
sebhmg, line 99: **suggestion (readability):** use pylint name instead of numbers:  `protected-access`

or better: fix the warning rather than disabling it

geoapps/driver_base/driver.py
```diff
@@ -97,6 +99,14 @@ def start(cls, filepath: str | Path, driver_class=None):
...
         )
 
+        version = ifile.data.get("version", None)
```
sebhmg, line 102: **suggestion:** cannot we guaranty it is backward compatible, and have the warning only if geoapps is older than the version in the file?

otherwise it seems users will get the warning often

sebhmg (commented):
is it realistic to ensure backward compatibility and warn only if geoapps package is older than version in the file?

In addition, it can still warn for older versions in file where backward compatibility was lost (that is versions that are known to be incompatible with current geoapps version).

geoapps/driver_base/driver.py
domfournier: That's old stuff and going to be deprecated. I ll just change the pylint name for now

geoapps/driver_base/driver.py
domfournier: Two things right now:
1- We have no mechanism to handle backward compatibility set yet.
2- Version is stored as a string with labels like 'alpha', 'rc', so we need to write/find code to interpret that.

Thoughts?

domfournier: Looks like these guys can handle string version comparison: https://packaging.pypa.io/en/latest/index.html

geoapps/driver_base/driver.py
```diff
@@ -96,7 +96,7 @@ def start(cls, filepath: str | Path, driver_class=None):
         print("Loading input file . . .")
         filepath = Path(filepath).resolve()
         ifile = InputFile.read_ui_json(
-            filepath, validations=driver_class._validations  # pylint: disable=W0212
+            filepath, validations=getattr(driver_class, "_validations", None)
```
sebhmg, line 99: ... looks like shoveling under the carpet. Disabling pylint is preferable as it is easier to track

geoapps/driver_base/driver.py
sebhmg: * about handling incompatibility: ideally, a new version should not introduce incompatibility with older ui.json. If it does and is not able to patch the ui.json to fix it, it should fail and report clearly about it. Out of scope here. Let's discussion live about it.
* for version comparison, this is semantic versioning: I would suggest using https://python-semver.readthedocs.io/en/latest/

sebhmg (changes requested):

geoapps/driver_base/driver.py
domfournier: Done

tests/version_test.py
```diff
@@ -34,10 +34,8 @@ def test_version_is_semver():
...
     oct_init["geoh5"] = str(assets_path() / "FlinFlon.geoh5")
     params = OctreeParams(**oct_init)
-    params.version = "0.0.0"
+    params.version = "10.0.0"
     params.write_input_file("test.ui.json", tmp_path)
 
-    with pytest.warns(
-        UserWarning, match="Input file version '0.0.0' does not match the."
-    ):
+    with pytest.warns(UserWarning, match="Input file version '10.0.0' is ahead"):
```
sebhmg, line 40: **suggestion:** have two other tests checking there is no warning for:
- current app version in file
- older app version in file, with something like:
```py
app_version = semver.Version.parse(geoapps.__version__)
if (app_version.minor > 0):
    version_in_file = app_version.replace(minor=app_version.minor-1)
else:
    version_in_file = app_version.replace(major=app_version.major-1)
```

tests/version_test.py
```diff
@@ -34,10 +34,8 @@ def test_version_is_semver():
 def test_input_file_version(tmp_path):
```
sebhmg, line 34: **suggestion (non-blocking):** could now use `semver.Version.is_valid()` above in `test_version_is_semver()`

sebhmg (changes requested):
thanks ! see suggestions in test

tests/version_test.py
```diff
@@ -18,24 +17,38 @@
...
         OctreeDriver.start(tmp_path / "test.ui.json")
+
+    if app_version.minor > 0:
+        version_in_file = app_version.replace(minor=app_version.minor - 1)
+    else:
+        version_in_file = app_version.replace(major=app_version.major - 1)
+
+    params.version = str(version_in_file)
+    params.write_input_file("test.ui.json", tmp_path)
+    OctreeDriver.start(tmp_path / "test.ui.json")
```
sebhmg, line 54: **suggestion:** ensure that no warnings are emitted

from [pytest doc](https://docs.pytest.org/en/7.4.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests):
```suggestion
    with warnings.catch_warnings():
        warnings.simplefilter("error")
        OctreeDriver.start(tmp_path / "test.ui.json")
```

sebhmg (changes requested):
in test, ensure that no warning is emitted

tests/version_test.py
domfournier: lovely

sebhmg (approved):

PR #608, opened 2023-09-01
GEOPY-1057: Change enforce_active to False

domfournier:
**GEOPY-1057 - Bad handling of topo from joint drivers**


domfournier:
@andrewg-mira @sebhmg  Weird pylint import-error here. Not sure if we have to tackle this here, but I can't reproduce it locally.

benk-mira (approved):

PR #611, opened 2023-09-20
GEOPY-1090: Fix deprecation warnings in tests

domfournier:
**GEOPY-1090 - Fix pytest deprecation warning for returning values in inversion runtests**


geoapps/utils/write_default_uijson.py
```diff
@@ -163,9 +163,17 @@ def write_default_uijson(path: str | Path, use_initializers=False):
     tipper_init = tipper_init if use_initializers else {}
 
     from geoapps.octree_creation.constants import app_initializer as oct_init
-
-    oct_init["geoh5"] = str(assets_path() / "FlinFlon.geoh5")
-    oct_init = oct_init if use_initializers else {}
+    from geoapps.octree_creation.constants import template_dict
```
benk-mira, line 166: How did this get in here?  I'm not opposed to adding three levels by default, but I think we had a single refinement by default to keep it consistent with Kris's ui.jsons.

tests/data_test.py
```diff
@@ -266,7 +266,7 @@ def test_normalize(tmp_path: Path):
     data.normalizations = data.get_normalizations()
     test_data = data.normalize(data.observed)
     assert np.all(
-        np.hstack(data.normalizations[None].values()).tolist()
+        np.hstack(list(data.normalizations[None].values())).tolist()
```
benk-mira, line 269:
```suggestion
        np.hstack(list(data.normalizations[None])).tolist()
```

tests/run_tests/driver_dc_2d_test.py
```diff
@@ -129,26 +127,18 @@ def test_dc_2d_run(tmp_path: Path, max_iterations=1, pytest=True):
...
         refinement=(4, 8),
     )
-
-    m_rec = test_dc_2d_run(
+    test_dc_2d_run(
         Path("./"),
         max_iterations=20,
         pytest=False,
     )
-    residual = np.linalg.norm(m_rec - m_start) / np.linalg.norm(m_start) * 100.0
```
benk-mira, line 150: How come?

benk-mira (changes requested):
Seems like some unrelated changes snuck in there.  See comment about compatibility with Kris's ui.jsons.

geoapps/utils/write_default_uijson.py
domfournier: Yea maybe, I will push back on Kris to get it full. I find it very annoying to have to modify the file every time. Don't think he will care since he has his own builder now...

tests/data_test.py
domfournier: Hmm not sure? `data.normalizations[None]` would technically be the keys, not the values?

tests/run_tests/driver_dc_2d_test.py
domfournier: It is mostly just visual inspection that we are doing anyway on the full run. We've never really enforced those values (except for gravity), and kind of a very loose metric, so why bother?

tests/data_test.py
benk-mira: oh, right.. sorry!

geoapps/utils/write_default_uijson.py
benk-mira: ok

benk-mira (approved):

PR #623, opened 2023-10-26
GEOPY-1132

domfournier:
**GEOPY-1132 - Jupyter-notebook area selection not applied**


geoapps/inversion/components/preprocessing.py
```diff
@@ -54,6 +67,9 @@ def window_data(
     :return: Updated data dict.
     :return: Vertices or centroids of the windowed data object.
     """
+    if isinstance(data_object, Grid2D):
```
MatthieuCMira, line 70: If we want the data_object to be points or subcalsses of points, shouldn't we verify it there?

tests/run_tests/apps_inversion_test.py
```diff
@@ -160,11 +161,14 @@ def test_mag_inversion(tmp_path: Path):
...
-        assert len(workspace.get_entity(app.params.data_object)) == 1
-        assert len(workspace.get_entity(app.params.mesh)) == 1
-        assert len(workspace.get_entity(app.params.topography_object)) == 1
+    filename = next(tmp_path.glob("MagneticVectorInversion_*.json"))
+    ifile = InputFile.read_ui_json(filename)
+    with ifile.data["geoh5"].open():
+        data = ifile.data["data_object"]
+        assert isinstance(data, Points)
+        assert data.n_vertices == 418
+        assert ifile.data["mesh"] is not None
```
MatthieuCMira, line 170: maybe test further than just is not None?
To verify the obejcts are preserved/transformed the right way?

MatthieuCMira (commented):
Minor comments,
Seems okay, I'll test it when review is done :-)

tests/run_tests/apps_inversion_test.py
domfournier: Fair enough

geoapps/inversion/components/preprocessing.py
domfournier: the data_object can be sub-class of BaseObject. We just need to do something special for Grid2D since it doesn't downsample as easily as vertex based objects. The inversion would go through, but were getting a lot of points with Nan at the end.

MatthieuCMira (approved):
Ok for me

MatthieuCMira (approved):

PR #624, opened 2023-10-27
GEOPY-1116

domfournier:
**GEOPY-1116 - update Pillow as recommended by GitHub advisory**


sebhmg:
see https://github.com/MiraGeoscience/geoapps/pull/625

PR #627, opened 2023-11-01
GEOPY-378

jamiefb:
**GEOPY-378 - Allow the ocTree mesh created with geoapps to be compatable with UBC**


geoapps/octree_creation/constants.py
```diff
@@ -99,6 +99,12 @@
             "main": True,
             "value": 500.0,
         },
+        "diagonal_balance": {
+            "group": "Basic",
+            "label": "UBC Compatible",
+            "main": True,
+            "value": False,
```
domfournier, line 106:
```suggestion
            "value": True,
```

I think we can default to True. You might have to turn it off for the unittests though to avoid changing the inversion runs.

geoapps/octree_creation/driver.py
```diff
@@ -116,6 +120,7 @@ def refine_tree_from_curve(
         mesh: TreeMesh,
         curve: Curve,
         levels: list[int] | np.ndarray,
+        diagonal_balance: bool,
```
domfournier, line 123: Defaulting to True here too is fine. Looks like you are already using a kwarg later in the code anyway.

domfournier (commented):
Great job. See small comments.

domfournier (approved):
Looking great. Moving to test

domfournier (approved):

PR #645, opened 2024-01-02
GEOPY-1265

domfournier:


geoapps/inversion/components/models.py
```diff
@@ -115,28 +115,31 @@ def active_cells(self, active_cells):
 
     @property
     def starting(self):
-        mstart = self._starting.model
+        mstart = self._starting.model.copy()
 
         if mstart is not None and self.is_sigma:
+            mstart = mstart.copy()
```
benk-mira, line 121: do you need both the copies?

benk-mira (approved):
looks like maybe a duplicate copy operation in there, but otherwise looks good

benk-mira (dismissed):

PR #647, opened 2024-01-11
GEOPY-1285: releasing 0.11.0

andrewg-mira:
**GEOPY-1285 - release full versions of geoh5py, param-sweeps, omf, las, geoapps**


sebhmg:
test failure, is due to randomness in data generation. Code in package is correct, but the test itself need to be fixed. Later. See GEOPY-1252

PR #654, opened 2024-01-29
GEOPY-1265

domfournier:
**GEOPY-1265 - Log bounds applied to twice on joint inversion**


geoapps/inversion/components/models.py
```diff
@@ -114,71 +114,73 @@ def active_cells(self, active_cells):
...
         if mstart is not None and self.is_sigma:
-            mstart = mstart.copy()
             mstart = np.log(mstart)
 
         return mstart
 
     @property
-    def reference(self):
+    def reference(self) -> np.ndarray | None:
         mref = self._reference.model
```
benk-mira, line 130: The logic in here is complicated.  Is there a way to combine the `if mref is None` and `if self.is_sigma & (all(mref == 0))`  and can we stick with one name for reference model?

benk-mira (changes requested):
Not clear how the diff solves the described issue..  I'll check during test.

benk-mira:
Whoops, I had the wrong issue in mind when I submitted my review!  ignore the comment about the diff not making sense!

geoapps/inversion/components/models.py
domfournier: You got it.

benk-mira (approved):

PR #658, opened 2024-02-06
GEOPY-1275

domfournier:
**GEOPY-1275 - DC forward fails on line id size.**


geoapps/inversion/electricals/driver.py
```diff
@@ -0,0 +1,150 @@
...
+from geoapps.inversion.components.windows import InversionWindow
+from geoapps.inversion.line_sweep.driver import LineSweepDriver
+from geoapps.inversion.params import BaseParams
+from geoapps.shared_utils.utils import get_locations, weighted_average
+from geoapps.utils.models import get_drape_model
+from geoapps.utils.surveys import extract_dcip_survey
+
+
+class BasePseudo3DDriver(LineSweepDriver):
+    _params_class: type(BaseParams)
```
benk-mira, line 36: mypy won't accept BaseParams as without type()?

geoapps/inversion/electricals/direct_current/pseudo_three_dimensions/driver.py
```diff
@@ -27,120 +16,10 @@
...
-from geoapps.utils.models import get_drape_model
-from geoapps.utils.surveys import extract_dcip_survey
+from geoapps.inversion.electricals.driver import BasePseudo3DDriver
 
 
-class DirectCurrentPseudo3DDriver(LineSweepDriver):
+class DirectCurrentPseudo3DDriver(BasePseudo3DDriver):
     _params_class = DirectCurrentPseudo3DParams
+    _params_2d_class = DirectCurrent2DParams
     _validations = validations
```
benk-mira, line 25: Why don't we need a super().__init__(params) in here?

geoapps/inversion/line_sweep/driver.py
```diff
@@ -116,19 +117,31 @@ def line_files(path: str | Path):
...
         data = {}
         drape_models = []
-        for line in lines:
+        for line in np.unique(line_ids):
             with Workspace(f"{path / files[line]}.ui.geoh5") as ws:
-                survey = ws.get_entity("Data")[0]
-                data = self.collect_line_data(survey, data)
-
-                mesh = ws.get_entity("Models")[0]
+                out_group = [
```
benk-mira, line 125: what if there is more than one SimPEGGroup?  Is there a ga_group_name we can use here?

geoapps/inversion/line_sweep/driver.py
```diff
@@ -186,16 +199,16 @@ def collect_results(self):
...
+                data[child.name] = {"values": np.zeros_like(line_indices) * np.nan}
 
-        ind = None
-        for child in survey.children:  # fill a chunk of values from one line
+        for child in survey.children:
             if "Iteration" in child.name:
-                if ind is None:
-                    ind = ~np.isnan(child.values)
-                data[child.name]["values"][ind] = child.values[ind]
+                data[child.name]["values"][line_indices] = child.values
```
benk-mira, line 212: won't this explode the number of entities?  Seems like you'll get an Iteration 0, 1, 2, ... for one line, then an Iteration 0 (0), 1 (0), 2(0), ... for the next line, and so on??

geoapps/inversion/params.py
```diff
@@ -346,6 +347,16 @@ def mesh(self):
     @mesh.setter
     def mesh(self, val):
         self.setter_validator("mesh", val, fun=self._uuid_promoter)
+
+        if (
```
benk-mira, line 351: Ah, ok - so this is why you removed all the data rotation stuff!

geoapps/utils/testing.py
```diff
@@ -140,50 +196,7 @@ def topo_drape(x, y):
...
-        survey = PotentialElectrode.create(
-            geoh5,
-            name="survey",
-            vertices=mn_vertices,
-            cells=np.vstack(dipoles).astype("uint32"),
-        )
-        survey.current_electrodes = currents
-        survey.ab_cell_id = np.asarray(current_id).astype("int32")
-        currents.potential_electrodes = survey
+        survey = generate_dc_survey(geoh5, X, Y, Z)
```
benk-mira, line 199: Great!  Starting to break this monster apart!

benk-mira (commented):
I'm not sure I fully understand the PR so my comments might not be that useful.. I'm sure it'll make more sense when I test!

geoapps/inversion/components/locations.py
```diff
@@ -33,12 +32,6 @@ class InversionLocations:
     ----------
     mask :
         Mask that stores cumulative filtering actions.
-    origin :
```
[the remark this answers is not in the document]
domfournier: We were blocking the process on rotated mesh already on the data: https://github.com/MiraGeoscience/geoapps/blob/330dc13c15189e5d7d2fc7bc137e1f3de16ff9cf/geoapps/inversion/components/data.py#L116

I think we just never cleaned things up all the way.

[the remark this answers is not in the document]
domfournier: There are just too many side effects for data with components

geoapps/inversion/electricals/driver.py
domfournier: `_param_class` store the type of Params, not the instantiated class.

geoapps/inversion/line_sweep/driver.py
domfournier: I see your point, but this is technically all internal mechanics of LineSweeper, so we already know that there is only going to be one group in there. We are not changing the name of that group as far as I can tell.

geoapps/inversion/line_sweep/driver.py
domfournier: I don't think so? This line would just be inserting values in the full array for every iteration it has. All lines share the same data iteration names, so the dict just get updated.

geoapps/inversion/params.py
domfournier: Same comment as above, looks like we were already blocking the process down the road. I am just formalizing it here. We would need to make serious push on all inversion methods to handle the rotation properly.

domfournier (commented):
See replies.

geoapps/inversion/line_sweep/driver.py
benk-mira: we'll find out for sure during test!

geoapps/inversion/electricals/driver.py
benk-mira: oh, right type(BaseParams) is type class or something.. gotcha!

geoapps/inversion/electricals/induced_polarization/pseudo_three_dimensions/constants.py
```diff
@@ -161,7 +158,7 @@
     },
     "chargeability_channel_bool": True,
     "chargeability_channel": {
-        "association": ["Cell", "Vertex"],
```
benk-mira, line 164: Testing this on the FlinFlon_dcip.geoh5 - I couldn't pick the Line ID object saved in the DC Survey unless I restored the "vertex"

benk-mira (commented):
Found an issue with the ui.json during testing

geoapps/inversion/electricals/induced_polarization/pseudo_three_dimensions/constants.py
domfournier: Means we need to update the example.

PR #671, opened 2024-04-08
GEOPY-1445: sync scripts and doc with version from the template repo

sebhmg:
**GEOPY-1233 - apply latest version of python-conda-template to all repos**
apply new name convention for lock files


sebhmg:
See instead https://github.com/MiraGeoscience/geoapps/pull/675

PR #673, opened 2024-04-16
GEOPY-1245 - update kmeans to ensure the clusters are always the same.

MatthieuCMira:
**GEOPY-1245 - Clustering: Group IDs in application don't match output**
The issue cames from the order of the cluster, which was random even if the seed was fixed. ordering the ids of the clusters based on their values fix the issue.


geoapps/clustering/driver.py
```diff
@@ -234,6 +251,9 @@ def run(self):
                 names=["Value", "Red", "Green", "Blue", "Alpha"],
             )
 
+            print(cluster_values)
+            print(group_map)
+
```
domfournier, line 256:
```suggestion
```

domfournier (approved):
Great idea. Small clean up

PR #677, opened 2024-04-22
GEOPY-1252: Fix random failure of tests

domfournier:
**GEOPY-1252 - Fix random failure of tests**


geoapps/utils/testing.py
```diff
@@ -491,7 +491,7 @@ def check_target(output: dict, target: dict, tolerance=0.1):
             / target["data_norm"],
             tolerance,
         )
-
+    print(output["phi_m"], target["phi_m"])
```
benk-mira, line 494: cleanup?  Or, it seems like this could be useful - maybe we should add target information to the top print statement?

benk-mira (changes requested):
update print in check_target to show target and output for each of data norm, phid and phim

geoapps/utils/testing.py
domfournier: Done

PR #679, opened 2024-04-23
GEOPY-1471: Inversion app fails on gravity

domfournier:
**GEOPY-1471 - Inversion app fails on gravity**


benk-mira (approved):
looks fine

PR #680, opened 2024-04-24
GEOPY1223

benk-mira:


geoapps/inversion/__init__.py
domfournier, line 1: You can delete all those files.

domfournier (commented):
The scope of this issue envolves deleteing all inversion files except the 1D inversion. Also need to update the apps so that they point to simpeg_drivers instead.

domfournier (commented):
Constant files should go too?

domfournier (approved):
Fantastic. The Great Purge ... not the staline one.

PR #684, opened 2024-05-01
GEOPY-1223: fix environment

sebhmg:


benk-mira (changes requested):
Need to remove some files that are defined in simpeg-drivers

benk-mira (approved):

PR #697, opened 2024-05-30
GEOPY-998: Re-route geoapps to Peak Finder standalone repo

domfournier:
**GEOPY-998 - Re-route geoapps to Peak Finder standalone repo**


geoapps/base/application.py
```diff
@@ -465,15 +465,54 @@ def create_copy(self, _):
...
+                    if isinstance(value, ObjectBase):
+                        obj = new_workspace.get_entity(value.uid)[0]
+                        if obj is None:
+                            obj = value.copy(parent=new_workspace, copy_children=True)
+                        param_dict[key] = obj
+
+            if self.live_link.value:
+                param_dict["monitoring_directory"] = self.monitoring_directory
+
+        new_params = type(self.params)(**param_dict)
```
benk-mira, line 513:
```suggestion
        new_params = self._param_class(**param_dict)
```

benk-mira (changes requested):
one small suggestion

geoapps/base/application.py
domfournier: The base class doesn't have `self._param_class`

PR #698, opened 2024-05-31
Geopy 1485

benk-mira:


geoapps/contours/application.py
```diff
@@ -9,50 +9,74 @@
...
-                self.defaults[key] = value.uid
+        if ui_json is not None and Path(ui_json).exists():
+            defaults = InputFile.read_ui_json(ui_json).data
+
+        if not defaults:
+            if Path(INITIALIZER["geoh5"]).exists():
+                defaults = INITIALIZER.copy()
             else:
-                self.defaults[key] = value
+                defaults = {}
```
domfournier, line 74: Shouldn't need this line.

geoapps/contours/application.py
```diff
@@ -125,28 +159,21 @@ def main(self):
...
-                )
-            )
-
     def update_name(self, _):
         if self.data.value is not None:
             self.export_as.value = self.data.uid_name_map[self.data.value]
         else:
             self.export_as.value = "Contours"
 
+    def is_computational(self, attr):
```
domfournier, line 168: Looks like the method is shared with edge_detection app. Consider moving it to the base class PlotSelection2D

domfournier: After looking at it more closely, this call is going to be very expensive as it generates an input file for every parameter in the loop of `get_param_dict`. Why does it have to access the fields of input file?

geoapps/contours/application.py
```diff
@@ -9,50 +9,74 @@
...
-                for key, value in app_initializer.items():
-                    if isinstance(value, uuid.UUID):
-                        app_initializer[key] = None
+        defaults = {}
 
-                self.params = self._param_class(**app_initializer)
+        if isinstance(geoh5, str):
+            if Path(geoh5).exists():
+                # defaults = {"geoh5": geoh5}
+                defaults = dict(INITIALIZER, **{"geoh5": geoh5})
```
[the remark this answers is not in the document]
benk-mira: yeah, good catch.  I rejigged the construction and fixed the run test

geoapps/contours/application.py
benk-mira: true.  removed.

geoapps/contours/application.py
benk-mira: I fixed the efficiency problem by moving INITIALIZER into the class and using that instead of the ui.json file.  I was about to move is_computational into the base class when I remembered that I already put it there to handle the old params class format.  So the child classes that use BaseData params override the definition for use with their params classes.

domfournier (approved):

PR #723, opened 2025-01-15
GEOPY-1873: geoapps tests are breaking after updating to latest published versions

domfournier:
**GEOPY-1873 - geoapps tests are breaking after updating to latest published versions**


benk-mira (approved):
Looks good to me.

PR #726, opened 2025-02-10
GEOPY-2007: Update geoapps with latest release dependencies

domfournier:
**GEOPY-2007 - Update geoapps with latest release dependencies**


geoapps/inversion/base_inversion_application.py
```diff
@@ -1652,17 +1640,35 @@ def write_trigger(
...
+                height=window_height,
+                azimuth=0.0,
+            )
+
+            if receivers_radar_drape is not None:
+                receivers_radar_drape = self.workspace.get_entity(
+                    uuid.UUID(receivers_radar_drape)
+                )[0]
+
+            drape_options = DrapeOptions(
```
benk-mira, line 1656: Hm.  If we keep supporting receiver shifting in the apps, do we need to keep the z_from_topo and other parameters in our simpeg_drivers params classes?  Won't this eventually try to feed z_from_topo into a params class in order to write the ui.json for the execution?

benk-mira (approved):

geoapps/inversion/base_inversion_application.py
domfournier: No, the opposite. This is meant to catch the geometric transformations (drape, window, etc) provided by the app, and do them before write to file.

PR #728, opened 2025-03-31
GEOPY-1163: update the readme of geoapps

domfournier:
**GEOPY-1163 - update the readme of geoapps**


domfournier:
> Ok with the text.
> 
> The installation seems complex though. No way to make it a no brainer bat? ;-)

That's the purpose of the "quick installation" section.

PR #730, opened 2025-04-14
[GEOPY-2095] Release/0.12.0

domfournier:
**GEOPY-2095 - Release geoapps-0.12.0**


sebhmg:
minor detail in the doc: navigation has the first item numbered, not the other ones.
Non-blocking, but I mention it here, @domfournier, in case you wish to fix it by the way, as you are already touching some doc files.
image

sebhmg (approved):

